PR #11, opened 2019-08-14
 Use '/crud/settings' api call to fetch layers instead of '/layer' and remove some mock 

dtrucs:


src/modules/CRUD/components/Nav/Nav.js
```diff
@@ -54,32 +41,43 @@ export const Nav = ({ getAllLayersAction, resizingMap, layersList, t }) => {
...
-      </ul>
+      { menu && (
+        <ul
+          id="CRUD-nav__menu"
+          className="CRUD-nav__menu"
+        >
+          {menu.sort(sortByOrder).map(({ name: group, pictogram: groupPictogram, crud_views: list }) => (
+            <li className="CRUD-nav__group" key={group}>
+              <p className="CRUD-nav__group-name">
+                {groupPictogram && (
```
mabhub: This block is used twice. Make it a component, and the rendering condition should goes inside it.

```jsx
const CrudNavIcon = ({ src, alt = '', ...rest }) => src && (
  <span className="bp3-icon" {...rest}>
    <img className="CRUD-nav__icon" src={src} alt={alt} />
  </span>
);
```

src/modules/CRUD/services/CRUD.js
```diff
@@ -0,0 +1,51 @@
+import Api from '@terralego/core/modules/Api';
+
+export async function fetchCRUDSettings () {
+  return Api.request('crud/settings');
+}
+
+export function getLayerFromCRUD ({ menu = [] }, name) {
```
mabhub: *function* is only for _Class methods_.

In any other case: _Arrow function_ should be preferred.


```suggestion
export const getLayerFromCRUD = ({ menu = [] }, name) => {
```

src/modules/CRUD/services/CRUD.js
```diff
@@ -0,0 +1,51 @@
...
+  const [view = {}] = menu.map(({ crud_views: views }) =>
+    views.find(item => item.layer.name === name)).filter(Boolean);
+
+  return view.layer ? { ...view.layer, niceName: view.name } : false;
+}
+
+export function getSourcesFromCRUD ({ menu = [] }) {
```
mabhub:
```suggestion
export const getSourcesFromCRUD = ({ menu = [] }) => {
```

src/modules/CRUD/services/CRUD.js
```diff
@@ -0,0 +1,51 @@
+import Api from '@terralego/core/modules/Api';
+
+export async function fetchCRUDSettings () {
```
mabhub:
```suggestion
export const fetchCRUDSettings = async () => {
```

src/modules/CRUD/services/CRUD.js
```diff
@@ -0,0 +1,51 @@
...
+          : [
+            ...sourceList,
+            { id: layer.group, type: 'vector', url: layer.group_tilejson },
+          ]
+      ), []),
+    ]
+  ), []);
+}
+
+export function getLayersPaintsFromCRUD ({ menu = [] }) {
```
mabhub:
```suggestion
export const getLayersPaintsFromCRUD = ({ menu = [] }) => {
```

src/modules/CRUD/components/Nav/Nav.js
dtrucs: Good point, done

src/modules/CRUD/components/Nav/Nav.js
```diff
@@ -54,32 +42,35 @@ export const Nav = ({ getAllLayersAction, resizingMap, layersList, t }) => {
...
-            </ul>
-          </li>
-        ))}
-      </ul>
+      { menu && (
+        <ul
+          id="CRUD-nav__menu"
+          className="CRUD-nav__menu"
+        >
+          {menu.sort(sortByOrder).map(({ name: group, pictogram: groupPictogram, crud_views: list }) => (
```
mabhub: It's not a good idea to `.sort()` array on each component render...

You should store sorted arrays as state that is recompute only on `menu` prop change.
Maybe something like:

```js
const [orderedMenu, setOrderedMenu] = useState(menu);

useEffect(() => {
  setOrderedMenu(menu.sort(SortByOrder));
  // but you also have menu sub items (`crud_views`) to manage...

  // You could propably extract such a sorting function into an helper:
  setOrderedMenu(getProperlyOrderedMenu(menu));
}, [menu]);
```

*...then menu would be processed only when `menu` prop is modified, and not on each render.*

dtrucs: ok, it's now fixed.
I added a filter to exclude a group of layer we don't need in the menu

src/modules/CRUD/components/Nav/Nav.js
```diff
@@ -1,39 +1,39 @@
...
-import { Button, Icon } from '@blueprintjs/core';
+import { Button } from '@blueprintjs/core';
 import classnames from 'classnames';
 
 import { generateURI } from '../../config';
+import NavIcon from './NavIcon';
 
 import './styles.scss';
 
+function sortByOrder ({ order: orderA }, { order: orderB }) {
```
mabhub:
```suggestion
const sortByOrder = ({ order: orderA }, { order: orderB }) => {
```

src/modules/CRUD/components/Nav/Nav.js
```diff
@@ -1,39 +1,39 @@
...
-          list: [
-            ...listItem.list,
-            item,
-          ],
-        }));
-    }
-    return [...list, { group: item.group, list: [item] }];
-  }, []);
+    setOrderedMenu(getFilteredAndOrderedMenu(menu));
+  }, [menuOpen, menu]);
```
mabhub: I think these should be in two separate `useEffect` hooks.

src/modules/CRUD/components/Nav/Nav.js
```diff
@@ -1,39 +1,41 @@
...
-          ...listItem,
-          list: [
-            ...listItem.list,
-            item,
-          ],
-        }));
-    }
-    return [...list, { group: item.group, list: [item] }];
-  }, []);
+  useEffect(() => {
```
hadrienl: One line it

src/modules/CRUD/components/Nav/Nav.js
```diff
@@ -54,32 +56,35 @@ export const Nav = ({ getAllLayersAction, resizingMap, layersList, t }) => {
...
-                  </NavLink>
-                </li>
-              ))}
-            </ul>
-          </li>
-        ))}
-      </ul>
+      { orderedMenu && (
+        <ul
+          id="CRUD-nav__menu"
```
hadrienl: Why this id ?

src/modules/CRUD/components/Details/Edit/index.js
```diff
@@ -17,7 +19,7 @@ export default withRouter(
     };
   })(
     connectCRUDProvider(({
-      layersList,
+      CRUDSettings,
```
hadrienl: Why `CRUDSetting`, capitalised, like a class. We are in CRUD context, so a single `settings` should be enough to know this is the CRUD'settings.

src/modules/CRUD/services/CRUD.js
```diff
@@ -0,0 +1,47 @@
+import Api from '@terralego/core/modules/Api';
+
+export const fetchCRUDSettings = async () => Api.request('crud/settings');
```
hadrienl: `async` is useless while you don't `await` anything.

src/modules/CRUD/services/CRUDProvider.js
```diff
@@ -39,17 +40,17 @@ export class CRUDProvider extends React.Component {
     }
   };
 
-  getAllLayersAction = async () => {
+  getCRUDSettings = async () => {
     try {
-      const { results: layersList = [] } = await fetchAllLayers();
-      this.setState({ layersList });
+      const CRUDSettings = await fetchCRUDSettings();
```
hadrienl: `settings`

src/modules/CRUD/services/CRUDProvider.js
```diff
@@ -39,17 +40,17 @@ export class CRUDProvider extends React.Component {
     }
   };
 
-  getAllLayersAction = async () => {
+  getCRUDSettings = async () => {
```
hadrienl: `getSettings`

src/modules/CRUD/components/Nav/Nav.js
dtrucs: It's for accessibility purpose, 
tldr: it's related to `aria-controls="CRUD-nav__menu"`

src/modules/CRUD/components/Nav/Nav.js
```diff
@@ -1,39 +1,36 @@
...
-          ...listItem,
-          list: [
-            ...listItem.list,
-            item,
-          ],
-        }));
-    }
-    return [...list, { group: item.group, list: [item] }];
-  }, []);
+  useEffect(() => { resizingMap(); }, [menuOpen]);
```
mabhub:
```suggestion
  useEffect(() => resizingMap(), [menuOpen]);
```

src/modules/CRUD/components/Nav/Nav.js
```diff
@@ -1,39 +1,36 @@
...
+  useEffect(() => { setOrderedMenu(getFilteredAndOrderedMenu(menu)); }, [menu]);
```
mabhub:
```suggestion
  useEffect(() => setOrderedMenu(getFilteredAndOrderedMenu(menu)), [menu]);
```

src/modules/CRUD/services/CRUD.js
```diff
@@ -0,0 +1,50 @@
+import Api from '@terralego/core/modules/Api';
+
+export const fetchSettings = () => Api.request('crud/settings');
+
+export const getLayerFromCRUD = ({ menu = [] }, name) => {
```
hadrienl: Is it really usefull to specify `fromCRUD` everywhere? We are in CRUD service scope, we should know that.

PR #21, opened 2019-08-27
fix #20 Load datalayer config from api instead of mocking

### LINKED ISSUE #20: Load datalayer config from backend
Today, the view list is hardcoded. We want to load this list from the backend to have a dynamic admin instead of a static one. This way, if we modify the backend configuration, the admin reflect those modifications.

jrmi:
- Load config from api in getDatalayerConfig (renamed getViewList)
- Create an HOC to inject this config as a props
- Update views that use this configuration
- avoid to load too many request by caching the value (need force update later)
- Rename datalayer config in view list


src/modules/DataLayer/services/datalayer.js
```diff
@@ -1,6 +1,19 @@
...
-// Mock api to return the config
-export const fetchDatalayerConfig = () => config;
+let datalayerconfig = null;
 
+export const fetchDatalayerConfig = async () => {
+  if (datalayerconfig === null) {
+    const apiViews =  await Api.request('geolayer/view/');
+
+    datalayerconfig = Object.entries(apiViews)
+      .map(elt => {
```
mabhub:
```suggestion
      .map([key, { pk: id, ...rest }] => ({ ...rest, id, key }))
```

src/modules/DataLayer/services/datalayer.js
```diff
@@ -1,6 +1,19 @@
...
+    // to frontend one.
+    // Should be removed in a future release
+    viewList = Object.entries(apiViews)
+      .map(([key, { pk: id, ...rest }]) => ({ ...rest, id, key }))
+      .sort(({ id: idA }, { id: idB }) => idA - idB);
+  }
+
+  return viewList;
+};
+export default { fetchViewList };
```
dtrucs: Why don't you use a provider and store `viewList` in a state ?
Then you can access to this viewList object everywhere with connecting the provider as HOC

mabhub: In a way setting up this HOC to fetch the list of available views is a temporary workaround to avoid storing _hard_ data in app.

In a near future, this `viewList` will be replaced by a react-admin resource (and a `<ReferenceInput />`). So there is no need to store in a global state that would not have any utility then.

jrmi: @dtrucs are you ok with that ?

PR #26, opened 2019-09-03
Improve and test CRUD services

dtrucs:


src/modules/CRUD/services/CRUDProvider.js
```diff
@@ -47,41 +50,73 @@ export class CRUDProvider extends React.Component {
...
 
   getFeaturesList = async layerId => {
+    if (!layerId) {
+      return;
+    }
     try {
-      const { results: featuresList = [] } = await fetchFeaturesList(layerId);
-      this.setState({ featuresList });
+      const { results: featuresList } = await fetchFeaturesList(layerId);
+      this.setState(state => ({
```
mabhub: As seen together, the returned object could be extracted in a small pure function, and maybe reused in cache block.

*And the best would be to have a way to call `setState` after the catch block (to avoid unexpected silent errors)*

src/modules/CRUD/services/CRUDProvider.js
```diff
@@ -29,105 +34,129 @@ export class CRUDProvider extends React.Component {
...
-      this.setState(state => ({
-        ...state,
-        errors: { ...state.errors, [featureId]: e.message },
-      }));
-      return null;
-    }
+    const { feature, error } = await saveFeatureAction(layerId, featureId, data);
+    const { errors, featuresList: prevFeaturesList } = this.state;
+
+    const isFeatureAlreadyExists = prevFeaturesList.some(({ identifier }) => (
```
mabhub:
```suggestion
    const doFeatureAlreadyExists = prevFeaturesList.some(({ identifier }) => (
```

```suggestion
    const isFeatureAlreadyExisting = prevFeaturesList.some(({ identifier }) => (
```

package.json
```diff
@@ -55,6 +55,12 @@
     "not ie <= 11",
     "not op_mini all"
   ],
+  "jest": {
+    "collectCoverageFrom": [
+      "src/**/*.js",
+      "!src/**/index.js"
```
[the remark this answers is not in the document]
dtrucs: Removed.

src/modules/CRUD/services/CRUDProvider.js
```diff
@@ -29,105 +34,123 @@ export class CRUDProvider extends React.Component {
...
+        propsFromIds.every(prop => item[prop] !== ids[prop])
+      ));
     }
+
+    return [...errorStore, error];
   }
 
+  getFeaturesList = async layerId => {
+    const { featuresList = [], error = {} } = await fetchFeaturesList(layerId);
+    const { errors } = this.state;
```
hadrienl: In the setState

src/modules/CRUD/services/CRUDProvider.js
```diff
@@ -29,105 +34,123 @@ export class CRUDProvider extends React.Component {
...
-      this.setState(state => ({
-        ...state,
-        errors: { ...state.errors, [featureId]: true, code: e.message },
-      }));
-    }
+    const { feature = {}, error = {} } = await fetchFeatureAction(layerId, featureId);
+
+    this.setState(state => ({
+      feature: (!feature.identifier)
+        ? { ...state.feature }
```
hadrienl: Destructure in function attribute.

src/modules/CRUD/services/CRUDProvider.js
```diff
@@ -29,105 +34,123 @@ export class CRUDProvider extends React.Component {
...
-      return feature;
-    } catch (e) {
-      this.setState(state => ({
-        ...state,
-        errors: { ...state.errors, [featureId]: e.message },
-      }));
-      return null;
-    }
+    const { feature = {}, error = {} } = await saveFeatureAction(layerId, featureId, data);
+    const { featuresList: prevFeaturesList } = this.state;
```
hadrienl: In setState

src/modules/CRUD/services/CRUDProvider.js
```diff
@@ -29,105 +34,123 @@ export class CRUDProvider extends React.Component {
...
-    try {
-      const settings = await fetchSettings();
-      this.setState({ settings });
-    } catch (e) {
-      this.setState(state => ({
-        ...state,
-        errors: { ...state.errors, code: e.message },
-      }));
-    }
+    const { settings = {}, error } = await fetchSettings();
```
hadrienl:
```suggestion
    try {
      const settings = await fetchSettings();
      this.setState({ settings });
      return settings;
    } catch (e) {
      this.setState({ settings: undefined, error: e.message });
      throw e;
    }
```

src/modules/CRUD/components/Details/index.js
```diff
@@ -11,18 +11,23 @@ export default withRouter(
...
-  }) => ({
-    map,
-    fetchFeature,
-    feature,
-    layer: getLayer(settings, layer),
-    hasError: errors[id],
-    errorCode: errors.code,
-  }))(
+  }) => {
+    const { error = {} } = errors.feature.find(({ featureId }) => featureId === id) || {};
```
mabhub:
```js
    const {
      // `error` can be destructured twice : first for having it entirely, then for its properties
      error = {},
      error: { message = '' } = {},
    } = errors.feature.find(({ featureId }) => featureId === id) || {};

    // Using an intermediate const for `feature` allow a more readable return{};
    const feature = featuresList.find(({ identifier }) => (identifier === id)) || {};
    
    return {
      map,
      fetchFeature,
      feature,
      layer: getLayer(settings, layer),

      // As you have defined a default value for undefined `error`, could it really be falsy?
      hasError: !!error,
      errorCode: message,
    };
```

src/modules/CRUD/services/CRUDProvider.js
```diff
@@ -29,105 +33,185 @@ export class CRUDProvider extends React.Component {
...
+      error,
+      ids,
+    } = props;
+    const { errors: { [store]: errorStore } } = this.state;
+
+    const propsFromIds = Object.keys(ids);
+
+    // Remove error if it was before in the store
+    if (!(error instanceof Error)) {
+      return errorStore.filter(item => (
```
mabhub: What kind of *item* are stored in `errorStore`? Wouldn't, for example, `error` be more meaningful than `item`?

src/modules/CRUD/services/CRUDProvider.js
```diff
@@ -29,105 +33,185 @@ export class CRUDProvider extends React.Component {
...
-        ...state,
-        errors: { ...state.errors, [state.featuresList.length]: true },
-      }));
+      result.error = e;
     }
-  }
+    const { featuresList = [], error = {} } = result;
+    const { errors } = this.state;
+
+    this.setState({
```
mabhub: `setState` is async ! you should use `errors` from `setState` callback argument.

src/modules/CRUD/services/CRUDProvider.js
```diff
@@ -29,105 +33,185 @@ export class CRUDProvider extends React.Component {
...
+
+    this.setState(({ errors }) => ({
+      featuresList,
+      errors: {
+        ...errors,
+        feature: this.getFormattedError({ error, ids: { layerId, featureId }, store: 'feature' }),
+      },
+    }));
+
+    return Object.keys(feature).length && feature;
```
mabhub: If length is `0`, `fetchFeature` will `return 0;`. Is that the right behavior?

src/modules/CRUD/services/CRUDProvider.js
```diff
@@ -29,105 +33,185 @@ export class CRUDProvider extends React.Component {
...
+    const isFeatureAlreadyExisting = prevFeaturesList.some(({ identifier }) => (
+      identifier === feature.identifier
+    ));
+
+    const featuresList = isFeatureAlreadyExisting
+      ? prevFeaturesList.map(item => (item.identifier === feature.identifier ? feature : item))
+      : [...prevFeaturesList, (feature.identifier) && feature].filter(Boolean);
+
+    this.setState(({ errors }) => ({
+      featuresList,
```
mabhub: `(prev)featuresList` should also be read from `setState` callback argument (like `error`), and not from a previous `this.state`.

Maybe `featuresList` should be defined using pure functions getting `prevFeaturesList` and `feature` as argument:

```js
this.setState(({  errors, featureList: prevFeatureList }) => ({
  featureList: onlyExistingFeatures(prevFeatureList, feature),
  errors: {
  // ...
```

src/modules/CRUD/services/CRUDProvider.js
```diff
@@ -29,105 +33,185 @@ export class CRUDProvider extends React.Component {
...
+      identifier === feature.identifier
+    ));
+
+
+    const featuresList = isFeatureAlreadyExisting
+      ? prevFeaturesList.map(item => (item.identifier === feature.identifier ? feature : item))
+      : [...prevFeaturesList, feature.identifier && feature].filter(Boolean);
+
+    this.setState(({ errors }) => ({
+      featuresList,
```
mabhub: Same as https://github.com/Terralego/terra-admin/pull/26#discussion_r325070614

src/modules/CRUD/services/CRUDProvider.js
dtrucs: Right, in this case the expected value is `false`. I fix it.

src/modules/CRUD/components/Details/index.js
dtrucs: In addition I changed `errorCode` by `errorMessage` that makes more sense.

src/modules/CRUD/services/CRUDProvider.js
mabhub: `!!` could have been the solution, but ok with yours.

src/modules/CRUD/services/CRUDProvider.js
```diff
@@ -29,105 +33,174 @@ export class CRUDProvider extends React.Component {
...
+          ? { error, ...ids }
+          : errorItem
+      ));
+    }
+
+    // Else add error in the store
+    return [...errorStore, { error, ...ids }];
+  }
+
+  onlyExistingFeatures = (featuresList, feature) => {
```
mabhub: `onlyExistingFeatures`is a _pure function_, it should be defined as a component method.

dtrucs: I rename this function with `updateOrSaveFeatureInFeaturesList` and now it's exported from `features.js`

PR #27, opened 2019-09-05
Add library + Add color picker field

DarkVisenya:
Waiting merge from https://github.com/Terralego/terra-layer/pull/4

### DIFF
--- a/public/locales/fr/translation.json
+++ b/public/locales/fr/translation.json
@@ -312,10 +312,10 @@
         "max-zoom": "Zoom maximum",
         "template": "Modèle de la fenêtre contextuelle"
       },
-      "minifiche": {
+      "minisheet": {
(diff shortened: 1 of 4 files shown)


public/locales/fr/translation.json
```diff
@@ -305,7 +305,8 @@
       },
       "minifiche": {
         "display-on-click": "Afficher la minifiche",
-        "template": "Modèle de la minifiche"
+        "template": "Modèle de la minifiche",
+        "highlight-color": "Choisir la couleur de surbrillance"
```
dtrucs: Can you add the verb that describes the action to the key?
Could you also change the key "minifiche" 3 lines above by a less french one?

DarkVisenya:
> @DarkVisenya Any reason for still being in Draft?

Yes, i cannot test this feature yet.

PR #37, opened 2019-09-17
Refacto and test Details and Edit components

dtrucs:


src/modules/CRUD/components/Details/Edit/Edit.js
```diff
@@ -70,18 +94,46 @@ class Edit extends React.Component {
...
+      feature: { geom = {} },
+      action,
+    } = this.props;
+
+    this.setState({
+      schema,
+      geom,
+    });
+    if (action !== ACTION_UPDATE || Object.keys(geom).length) {
+      this.initDraw();
```
mabhub: Shouldn't this be as [second argument of `setState`](https://fr.reactjs.org/docs/react-component.html#setstate)?

src/modules/CRUD/components/Details/Edit/index.js
```diff
@@ -31,8 +31,8 @@ export default withRouter(
         identifier === id
       )) || {},
       saveFeature,
-      layer: getLayer(settings, layer),
-      layerPaint: getLayersPaints(settings).find(item => item['source-layer'] === layer),
+      layer: getLayer(settings, layer) || {},
```
mabhub: Shouldn't the fallback value `{}` be provided by `getLayer()` function?

PR #55, opened 2019-09-24
Improve Read component (CSS and title)

dtrucs:


src/modules/CRUD/components/Details/Templates/Templates.js
```diff
@@ -2,8 +2,8 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { AnchorButton, ButtonGroup } from '@blueprintjs/core';
 
-const Templates = ({ files, id }) => id && files.length > 0 && (
-  <ButtonGroup>
+const Templates = ({ files, id, ...rest }) => id && files.length > 0 && (
```
mabhub: Shouldn't this component have a more agnostic name like:
- `DownloadButtonGroup`
- `DownloadButtons`
- `FileDownloadButtons` 

_'cause there's no business logic really concerning 'templates' or 'templating'._

PR #59, opened 2019-09-26
Adapt CRUD config with the API breaking change

dtrucs:


mabhub:
Could you explain "API breaking change"?

dtrucs:
@mabhub The API doesn't return the same JSON anymore:
- In the layer object, the key `tilejson` replaces `group_tilejson` (Cause it could be `group_tilejson` or `layer_tilejson`, the backend gives directly the needed value)
- Before we needed to know the `source` and now this keys was removed because we can use the `id`

PR #64, opened 2019-09-27
Override i18next language detection to french

mabhub:
...as we do not have any other language contributed for now

### DIFF
--- a/src/config/i18n.js
+++ b/src/config/i18n.js
@@ -10,6 +10,7 @@ i18n
   .use(LanguageDetector)
   .use(reactI18nextModule) // if not using I18nextProvider
   .init({
+    lng: 'fr',
     backend: {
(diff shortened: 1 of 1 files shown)


src/config/i18n.js
```diff
@@ -10,6 +10,7 @@ i18n
   .use(LanguageDetector)
   .use(reactI18nextModule) // if not using I18nextProvider
   .init({
+    lng: 'fr',
```
dtrucs: Why do you need to override it?
The fallbackLng below is not enough?

mabhub: If user browser locale is **en**, it is used as _detected language_, but we do not have english translations.

SebCorbin: > english translations are not fully contributed.

Are they not the default ones?

PR #65, opened 2019-09-27
Fetch feature item instead of searching in the current featureList

dtrucs:


src/modules/CRUD/services/CRUDProvider.js
```diff
@@ -139,17 +140,24 @@ export class CRUDProvider extends React.Component {
       result.error = e;
     }
 
-    const { feature = {}, error = {} } = result;
+    const { feature: newFeature = {}, error = {} } = result;
 
-    this.setState(({ errors, featuresList }) => ({
-      featuresList: updateOrSaveFeatureInFeaturesList(featuresList, feature),
+    this.setState(({ errors, feature, featuresList }) => ({
+      feature: {
```
mabhub: And what about an pretty `updateFeatureIdentifier()` function? :smiley: 

like:
```suggestion
      feature: updateFeatureIdentifier(feature, newFeature),
```

src/modules/CRUD/services/CRUDProvider.js
```diff
@@ -161,17 +169,24 @@ export class CRUDProvider extends React.Component {
...
```
mabhub: Same as above...

src/modules/CRUD/services/CRUDProvider.js
```diff
@@ -184,8 +199,9 @@ export class CRUDProvider extends React.Component {
     }
     const { feature } = result;
 
-    this.setState(({ featuresList }) => ({
-      featuresList: !feature
+    this.setState(({ featuresList, feature: { [feature]: deletedFeature, ...rest } }) => ({
+      feature: rest,
+      featuresList: !deletedFeature
```
mabhub: I see that's from previous code but could you invert the ternary:

```js
      featuresList: deletedFeature
        ? featuresList.filter(({ identifier }) => identifier !== featureId), 
        : featuresList
```

PR #73, opened 2019-10-02
Add hack to be able to save viewpoint while backend api is not updated

jrmi:
- Add hack to create viewpoints by adding missing data
- Merge geometry and endpoint
- Add index field
- Center map with the map config setting on viewpoint creation
- Use configMap mapboxToken to create mapboxgl map


src/components/react-admin/MapPointInput.js
```diff
@@ -46,11 +58,14 @@ const MapPointInput = ({ classes, input }) => {
...
       )}
-    </Map>
+    </Map>))
   );
 };
 
 export default compose(
   addField,
   withStyles(styles),
+  connectAppProvider(({ env: { configMap } }) => ({
```
mabhub: What about creating a small `withMapConfig` HOC to abstract provider connection complexity?

src/services/react-admin/patchDataProvider.js
```diff
@@ -0,0 +1,83 @@
...
...vAt+WK4s9lMng3NMOONTTbSAsClQrXZ2p51R/hnSB1xa1/LfkuyOZmTazNAwaY/BtTgnwmkzWNpi3p06fF369taIIFd+nS9G80CHIrUs2rvgjvBLmwoD9xOTpSkI/8/NDL1RUcx8Hbzc3sBKIpQaBSIb9vX3AcBy8XF3wlv9/0/Eoub/rbpwcOpEqQX0NUOHGeCWIz/mc22dGrjhDkF4UC12Qy1AkEUDQbWdr72GNmP8+cIJ/JZPBsECy5Z0/cUyqxuX9/cByHfm5u+K6ZcDQIckevQt3vmCA2489J5kNEqyBdXFwgdHeH0N0dg9zdm+YxGvFwccGuZr/0DyNIvUqFkd27N9VAX8vl0DSI8MwDwlEhSCATxKZ8kGI5SDQKYolpXl742M+vzc8zJwhUKlyTyZomHCMaZs0fc3fH1/KWo300CHI7SIU61sSyHe+l8U8Qgbs70nr2bGKCp2dTLeLbqRP+bmYew1pBoFJhopdXC/Gy+/RpNRrGBHECQf6YYXqUh2ZBTHXSlza8nuM4aDp3tjgbbo0grzYbKu7t6mry/WgQhHXSbczluXSMYmU9oiA3FAoMabaIMKxrV/xgYQl8W4LcUSqb+iI53t4m34MJ4gSCXDTQIUjjGidXjsPv2znMe0uhgH+zUaxFvXvjjhWLFU0JApUK38jl+JNE0mLkijZBfglV82rfCO8EeZeCeZC7SiWGdevWNAL1nqT1bkVr50E+9vNrsYy9oG9fq9ZiWbtZijpBwpggNuWdx/sSF+RLuRwDGpZx9HZ1xTUT+zgeZib9JaGwKbjdXFxQY2JFr6MI8vMQJohNeWux/QRZ178/tg8ciJd9fHBRIsE7YjH2DRrUJAfHccjs1cvk4sKHEaRepcK2AQOahmpdOA5vikSOKUg4E8Sm/O4J+wky1tPT4hxG83VOjyJII4aGTVgcx6GHqyv+r9k8hqMIciuCrcWyKeeW9rFbYc54YG6hEW83N2zo3x//lptfdt8eQW4qFJjS7DMVHh640dDhdhhB2GJF2/JGtv0EqW8I5mWJBGeEQpwSCvEniQT3CATLUfhpqJpXd5nwTpDXV/QmXsiMRxAkigliU15dyQThMzejmSA25eVVvYgXMuMRBIlR4/g7TBCb8eLqnsQLmdF+bgxjgtiU0wVMED5zYzi/BPl/9hKfH3RE1GkAAAAASUVORK5CYII=';
```
mabhub: Could you please use a lighter/smaller encoded picture to avoid add such a weight to commit history?


This _(`98 o`)_ blue square for example:

index-fs8


```suggestion
const defaultImage = 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAJ4AAACaAQMAAABi2mMuAAAAA1BMVEVHbJjlM5lCAAAAGklEQVRIx+3BMQEAAADCoPVPbQ0PoAAAAF4MDKIAAfsS1FsAAAAASUVORK5CYII=';
```

src/components/react-admin/MapPointInput.js
```diff
@@ -5,12 +5,12 @@ import { addField } from 'react-admin';
 // eslint-disable-next-line import/no-extraneous-dependencies
 import { withStyles } from '@material-ui/core/styles';
 
-import compose from '../../utils/compose';
+import { getSettings } from '../../services/settings';
```
mabhub: This `import` seems to be unused.

src/components/react-admin/MapPointInput.js
```diff
@@ -5,12 +5,12 @@ import { addField } from 'react-admin';
...
+import { connectAppProvider } from '../AppProvider';
 
+import compose from '../../utils/compose';
 
-const Map = ReactMapboxGl({
-  accessToken: 'pk.eyJ1IjoibWFraW5hY29ycHVzIiwiYSI6ImNqY3E4ZTNwcTFta3ozMm80d2xzY29wM2MifQ.Nwl_FHrWAIQ46s_lY0KNiQ',
-});
+let Map = null; // For the mapbox gl component
```
mabhub: As you load it from `useEffect` hook, maybe storing it into component state would be preferable, wouldn't it?

mabhub: *(Also loaded state wouldn't be required anymore as you could simply check if map (state) is truthy or not.)*

src/modules/RA/Viewpoint/components/ViewpointFields.js
```diff
@@ -56,16 +59,24 @@ const ViewpointFields = ({ edit, classes, ...props }) => {
   };
 
   React.useEffect(() => {
-    if (edit) {
+    if (edit || create) {
```
mabhub: As if edit OR create should be always truthy, condition id now useless. This could be simplfy to:

```js
React.useEffect(() => {
  getRemoteData();
}, []);
```

src/modules/RA/Viewpoint/views/Create.js
```diff
@@ -9,7 +9,7 @@ export const ViewpointCreate = props => (
     {...props}
     actions={<DefaultActions />}
   >
-    <ViewpointFields />
+    <ViewpointFields create />
```
mabhub: I don't see the point of this prop...

src/components/react-admin/MapPointInput.js
jrmi: Removed !

src/components/react-admin/MapPointInput.js
jrmi: I tried, but it failed inside ReactMapboxGL code. I think the map object is not suitable for use as a component state, but why ?

src/modules/RA/Viewpoint/views/Create.js
jrmi: Is create equivalent to !edit. In this case I agree.

src/components/react-admin/MapPointInput.js
jrmi: Agree, Done !

src/components/react-admin/MapPointInput.js
jrmi: We can see this later if you want.

src/modules/RA/Viewpoint/components/ViewpointFields.js
jrmi: Done

src/modules/RA/Viewpoint/views/Create.js
jrmi: Done

src/components/react-admin/MapPointInput.js
```diff
@@ -34,10 +33,21 @@ const MapPointInput = ({ classes, input }) => {
...
+    if (mapConfig.accessToken) {
+      Map = ReactMapboxGl({
+        accessToken: mapConfig.accessToken,
+      });
+      setLoaded(true);
+    }
+  }, [mapConfig.accessToken]);
+
   return (
+    (loaded && (
```
mabhub: There's a pair of extraneous parents:
You have `return (loaded && (<Map>...</Map>));` where `return loaded && (<Map>...</Map>);` would be enough.

Although, I would personaly prefer this :
```js
if (!loaded) {
  return null; // Or maybe `return <Loading />`;
}

return (
  <Map
    ...
  >
  ...
  </Map>
);
```

package.json
```diff
@@ -6,6 +6,7 @@
     "@blueprintjs/core": "~3.18.1",
     "@blueprintjs/datetime": "~3.13.0",
     "@blueprintjs/icons": "~3.10.0",
+    "@material-ui/core": "^4.5.0",
```
mabhub: On one hand you add depdency to `@material-ui/core` and on the other hand, you add some `// eslint-disable-line import/no-extraneous-dependencies` linting exception ;)
We have a choice to make!

I would personaly prefer not setting `@material-ui/core` as project dependency, as It imply sticking to a specific version, while the sole hard dependency is through react-admin. In case of upgrade of react-admin, it could result in a *desync* of version between our and the one required by react-admin. It's not a serious issue, but that was the starting point of not defining this dependency until now.

For now I do not know what is the best solution. But I'm sure that we should standardize in only one of these way:
- Keep eslint exception and not adding MUI as dependency, or:
- Keep dependency to MUI and remove all eslint exception that are concernd

public/locales/fr/translation.json
```diff
@@ -509,8 +510,8 @@
         "properties.keywords": "Mots-clés",
         "properties.landscape-entities": "Entités paysagères",
         "properties.related-elements": "Documents associés",
-        "geometry.coordinates[0]": "Longitude",
-        "geometry.coordinates[1]": "Latitude"
+        "point.coordinates[0]": "Longitude",
+        "point.coordinates[1]": "Latitude"
```
dtrucs: Maybe it's preferable to be more verbose such as :
"point.coordinates.lng": "Longitude",
"point.coordinates.lat": "Latitude"

src/components/react-admin/MapPointInput.js
```diff
@@ -34,10 +33,21 @@ const MapPointInput = ({ classes, input }) => {
     customProps.center = input.value;
   }
 
+  React.useEffect(() => {
+    if (mapConfig.accessToken) {
+      Map = ReactMapboxGl({
```
dtrucs: Sorry for the repost, I still don't understand why we use React Mapbox Gl instead of `Map` component of terralego/core which is also a wrapper of Mapbox Gl and it's already packaged in this project.

src/modules/RA/Viewpoint/components/ViewpointFields.js
```diff
@@ -146,19 +156,22 @@ const ViewpointFields = ({ edit, classes, ...props }) => {
...
         <FileField source="related" src="document" title="key" />
       </FormTab>
+      )}
 
+      {edit && (
       <FormTab label="resources.viewpoint.tabs.pictures" path="pictures">
         <ReferenceArrayField source="picture_ids" reference={RES_PICTURE} fullWidth>
           <Datagrid rowClick="edit">
-            <TextField source="owner" />
+            {false && <ReferenceField source="owner_id" reference="users" />}
```
dtrucs: Wut?

public/locales/fr/translation.json
mabhub: Back-end API provide only an array of coordinates and we are using it "as is".

src/modules/RA/Viewpoint/components/ViewpointFields.js
mabhub: In addition to commit title, I added a code comment to explain it.

src/components/react-admin/MapPointInput.js
mabhub: [Map component](https://github.com/Terralego/terra-front/blob/master/src/modules/Map/Map/Map.js) from [terra-front](https://github.com/Terralego/terra-front) does not allow _composition_ of MapboxGL elements. It only provides management for a custom configuration object which should provide all map content.

Current use case won't take any advantage of it.

PR #75, opened 2019-10-03
Handle related documents

SebCorbin:


src/modules/RA/Viewpoint/components/ViewpointFields.js
```diff
@@ -3,20 +3,22 @@ import PropTypes from 'prop-types';
...
   AutocompleteArrayInput,
   Datagrid,
   DateField,
   DisabledInput,
   EditButton,
-  FileField,
-  FormTab,
   ImageField,
+  ImageInput,
+  FormTab,
```
mabhub: Could you keep `FormTab` just before `ImageField`? :p

PR #79, opened 2019-10-07
Set fitBounds duration to 0 (no more animation)

dtrucs:


src/modules/CRUD/views/Map/Map.js
```diff
@@ -188,7 +188,7 @@ export class Map extends React.Component {
         left: 20,
       };
       map.resize();
-      map.fitBounds(getBounds(coords), { padding });
+      map.fitBounds(getBounds(coords), { padding, duration: 0 });
```
mabhub: Should this come from a prop with 0 as default value?

dtrucs: Right, now the value can be passed by a param

PR #90, opened 2019-10-10
Delete picture with redirect

jrmi:
When deleting a picture the redirect doesn't work and we come back to the list. So a new `<customToolbar />` customize all button of the form, including the delete button with a redirect.


src/components/react-admin/CustomToolbar.js
```diff
@@ -0,0 +1,55 @@
...
+    display: 'flex',
+    justifyContent: 'flex-start',
+  },
+  filler: {
+    flex: 1,
+  },
+};
+
+
+const PostCreateToolbar = ({
```
SebCorbin: Let's add some documentation

jrmi: You're so damn right ! Done. (Also the naming was copypasted).

PR #92, opened 2019-10-10
Rise the page size of featureList

dtrucs:
Waiting for the work of pagination to be done, this is a temporary fix.


src/modules/CRUD/services/features.js
```diff
@@ -1,7 +1,7 @@
 import Api from '@terralego/core/modules/Api';
 
 export const fetchFeaturesList = layerId =>
-  Api.request(`crud/layer/${layerId}/features/`);
+  Api.request(`crud/layer/${layerId}/features/?page_size=2000`);
```
jrmi: Should be a param with default to 2000 ?

dtrucs: As i said in the description of the PR, this is a temporary fix.

dtrucs: 2000 items per page.

PR #93, opened 2019-10-10
Add Table view in Read component

dtrucs:


src/modules/CRUD/components/Details/Read/Read.js
```diff
@@ -13,6 +13,28 @@ const NO_FEATURE = 'CRUD.details.noFeature';
...
+const isTableObject = value => (
+  value.reduce((list, item) => {
+    if (!list.columns.length) {
+      return {
+        ...list,
+        columns: Object.keys(item),
+        rows: [Object.values(item)],
+      };
+    }
+    if (!Object.keys(item).sort().every((prop, i) => list.columns.sort()[i] === prop)) {
```
mabhub: `if (whaaaaaats‽…) {`

Could you put this in a tiny pure function with a self explanatory name? _(which will probably take `item` & `list` as arguments)_


> « It's harder to read code than to write it »

src/modules/CRUD/components/Details/Read/Read.js
```diff
@@ -13,6 +13,28 @@ const NO_FEATURE = 'CRUD.details.noFeature';
 
 const emptyStringNullOrUndef = value => ['', null, undefined].includes(value);
 
+const isTableObject = value => (
+  value.reduce((list, item) => {
+    if (!list.columns.length) {
```
mabhub: I see three cases in this `reduce()` body. A simple line of comment above each case could easily explain them.

src/modules/CRUD/components/Details/Read/Read.js
```diff
@@ -13,6 +13,37 @@ const NO_FEATURE = 'CRUD.details.noFeature';
...
+
+    // Fill the columns and rows arrays with the first iteration
+    if (!list.columns.length) {
+      return {
+        ...list,
+        columns,
+        rows: [rows],
+      };
+    }
+    // Define the object as no table
```
mabhub:
```suggestion

    // If current iteration has different columns, it is not a table.
```

src/modules/CRUD/components/Details/Read/Read.js
```diff
@@ -13,6 +13,37 @@ const NO_FEATURE = 'CRUD.details.noFeature';
...
 
+const haveStrictlySameValues = (arr1, arr2) =>
+  [...arr1].sort().join() === [...arr2].sort().join();
+
+const isTableObject = value => (
+  value.reduce((list, item) => {
+    const columns = Object.keys(item);
+    const rows = Object.values(item);
+
+    // Fill the columns and rows arrays with the first iteration
```
mabhub:
```suggestion
    // If `list`does not have any columns (first iteration), fill array with current cols & rows
```

src/modules/CRUD/components/Details/Read/Read.js
```diff
@@ -13,6 +13,37 @@ const NO_FEATURE = 'CRUD.details.noFeature';
...
+      };
+    }
+    // Define the object as no table
+    if (!haveStrictlySameValues(columns, list.columns)) {
+      return {
+        ...list,
+        isTable: false,
+      };
+    }
+    // Fill rows of each iteration
```
mabhub:
```suggestion

    // Columns are the same: fill rows with current rows
```

dtrucs:
Following our discussion, everything was refacto differently

PR #96, opened 2019-10-11
Split DateTimeInput in two component DateInput & TimeInput

mabhub:


src/components/react-admin/TimeInput.js
```diff
@@ -0,0 +1,88 @@
...
+  alwaysOn, basePath, component, defaultValue, formClassName, initializeForm, input, isRequired,
+  label, limitChoicesToValue, locale, meta, options, optionText, optionValue, record, resource,
+  allowEmpty, source, textAlign, translate, translateChoice, ...rest
+}) => rest;
+
+const leftPad = (nb = 2) => value => ('0'.repeat(nb) + value).slice(-nb);
+
+const convertDateToTimeString = date => {
+  if (!(date instanceof Date) || Number.isNaN(date)) return '';
+  const hh = leftPad(2)(date.getHours());
```
dtrucs: Why don't you use `date.getHours().toString()`?

PR #114, opened 2019-10-25
Display viewpoint index instead of technical id in list

AL9000:


src/components/react-admin/GridList.js
```diff
@@ -46,6 +46,11 @@ const getColsForWidth = width => {
   return 6;
 };
 
+const getViewpointSubtitle = (index, t) => {
+  if (Number.isNaN(Number(index))) return t('resources.viewpoint.viewpoint-no-number');
+  return t('resources.viewpoint.viewpoint-number', { number: index });
+};
```
mabhub:
```js
const viewpointSubtitle = index => (
  Number.isNaN(Number(index))
    ? ['resources.viewpoint.viewpoint-no-number']
    : ['resources.viewpoint.viewpoint-number', { number: index }]
);
```

```js
<GridListTileBar
  className={classes.tileBar}
  title={data[id].label}
  subtitle={<span>{t(...viewpointSubtitle(data[id].properties.index))}</span>}
/>
```

PR #119, opened 2019-10-29
Improve display of fields error

dtrucs:


src/modules/CRUD/components/Details/Edit/ErrorListTemplate.js
```diff
@@ -2,22 +2,43 @@ import React from 'react';
...
-const ErrorListTemplate = ({ errorSchema: { __errors, ...fields }, schema: { properties } }) => {
-  const errors = Object.keys(fields).reduce((list, item) => {
-    const { __errors: errorsItem } = fields[item];
-    return errorsItem.length
-      ? [...list, { title: properties[item].title, list: errorsItem }]
-      : list;
-  }, []);
+const getProperties = props => Object.keys(props).reduce((list, item) => {
+  const { type, properties, title } = props[item];
+  if (type === 'object') {
```
mabhub: You could group this in a single `return`:

```js
  const customProperties = type === 'object'
    ? getProperties(properties)
    : { [item]: title };

  return {
    ...list,
    ...customProperties,
  };
```

PR #123, opened 2019-10-31
Improve schemaform style

dtrucs:


src/modules/CRUD/components/Details/Edit/Edit.js
```diff
@@ -34,6 +34,35 @@ const updateSchemaPropertiesValues = (properties, formData) => (
...
+            ...content.props,
+            uiSchema: {
+              ...content.props.uiSchema,
+              classNames: '',
+            },
+          },
+        };
+        return <div className={`col-xs-12 ${classNames}`}>{contentModified}</div>;
+      })}
+    </div>
```
mabhub:
```js
/**
 * extractClassnames
 *
 * >>> Testable pure function
 */
const extractClassnames = ({
  content,
  content: { props: { uiSchema: { classNames = '' } } },
  ...rest
}) => {
  const contentWithoutClassNames = {
    ...content,
    props: {
      ...content.props,
      uiSchema: {
        ...content.props.uiSchema,
        classNames: '',
      },
    },
  };

  return {
    content: contentWithoutClassNames,
    classNames,
    ...rest,
  };
};

/**
 * Another way to implement the same thing:
 */

const extractClassnames = property => {
  const classNames = _.get(property, 'content.props.uiSchema.classNames');

  if (classNames) {
    // eslint-disable-next-line no-param-reassign
    delete property.content.props.uiSchema.classNames;
  }

  return {
    ...property,
    classNames,
  };
};

```

```jsx
  <div className="row">
    {properties.map(property => {
      const { content, classNames } = extractClassnames(property);
      return <div className={`col-xs-12 ${classNames}`} key={someKey}>{content}</div>;
    })}
  </div>
```

mabhub: This could also be done as a HOC that will wrap `ObjectFieldTemplate` and give a modified `properties` directly as prop.

PR #155, opened 2019-12-03
DraggableFormIterator for layer fields ordering

SebCorbin:


package-lock.json
```diff
@@ -7356,25 +7356,25 @@
       "dependencies": {
         "abbrev": {
           "version": "1.1.1",
-          "resolved": "",
+          "resolved": false,
```
mabhub: `package-lock.json` should'nt have change so much..

src/components/react-admin/DraggableFormIterator.js
```diff
@@ -0,0 +1,282 @@
+/* eslint-disable
```
mabhub: Can you make a first commit including this copy of FormIterator with no change at all, and then another commit with the changes you made?

This would allow identifying your changes, and re-applying them in case of FormIterator original file upgrade.

src/modules/RA/DataLayer/components/SourceFetcher.js
```diff
@@ -23,12 +23,13 @@ const SourceFetcher = ({ dispatch, dataProvider, sourceId, fields = [] }) => {
...
     const fillFields = async () => {
       const { data: { _type: type, fields: sourceFields = [] } } = await load(sourceId);
-      const filledFields = sourceFields.map(({ id, name, label }) => ({
-        id,
-        name,
-        label,
-        ...fields.find(({ id: fieldId }) => id === fieldId) || {},
-      }));
+      const filledFields = [
+        ...sourceFields.filter(({ id }) => !fields.find(({ id: fieldId }) => id === fieldId)),
```
jrmi: Can you un-inline this `filter` and the next `map` with an explicit var name to enhance readability ?

src/components/react-admin/DraggableFormIterator.js
mabhub: Also, as in `/src/components/react-admin/CustomFormIterator.js`, please add a comment explaining in wich way the file diverge from original one.

SebCorbin: Added a comment to explain from what it has been implemented

SebCorbin: I did not start from SimpleFormIterator but from another PR https://github.com/marmelab/react-admin/pull/1836 so it has greatly diverged and it wouldn't mean anything

src/modules/RA/DataLayer/components/SourceFetcher.js
```diff
@@ -23,12 +23,19 @@ const SourceFetcher = ({ dispatch, dataProvider, sourceId, fields = [] }) => {
 
     const fillFields = async () => {
       const { data: { _type: type, fields: sourceFields = [] } } = await load(sourceId);
-      const filledFields = sourceFields.map(({ id, name, label }) => ({
-        id,
-        name,
-        label,
-        ...fields.find(({ id: fieldId }) => id === fieldId) || {},
+      const fieldsFromSource = sourceFields.filter(
+        ({ id }) => !fields.find(({ id: fieldId }) => id === fieldId),
```
mabhub: Using [`.some()`](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/Array/some) instead of `.find()` would return directly a boolean.

PR #172, opened 2019-12-17
Activate relevant filters and search on sources and layers

SebCorbin:
Depending on https://github.com/Terralego/terra-layer/pull/19 and https://github.com/Terralego/django-geosource/pull/7

### DIFF
--- a/src/modules/RA/DataLayer/views/List.js
+++ b/src/modules/RA/DataLayer/views/List.js
@@ -1,15 +1,17 @@
 import React from 'react';
 import {
-  List, Datagrid,
-  FunctionField,
-  TextField,
-  TextInput,
+  BooleanField,
+  BooleanInput,
+  Datagrid,
   EditButton,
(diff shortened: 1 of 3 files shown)


src/modules/RA/DataLayer/views/List.js
```diff
@@ -30,6 +32,29 @@ const ListFilters = props => (
...
+    >
+      <SelectInput />
+    </ReferenceInput>
+    <BooleanInput
+      source="active_by_default"
+      label="resources.datalayer.fields.active_by_default_pastpart"
+    />
+    <BooleanInput
+      source="table_enable"
+      label="resources.datalayer.fields.table_enable"
```
mabhub: react-admin provide a default label corresponding to the form element path _(which in this case is `resources.datalayer.fields.table_enable`)_. So you don't need to add manuel label if you do not require a different translation key.

```suggestion
```

PR #174, opened 2019-12-19
Add a recursion to correctly display all nested formats

dtrucs:


src/modules/CRUD/components/Details/Read/DefaultView/DefaultView.js
```diff
@@ -62,14 +62,14 @@ const formattedProp = ({ value, t }) => {
           <table className="details__table">
             <thead>
               <tr>
-                {columns.map(th => <th key={th}>{th}</th>)}
+                {columns.map(th => <th key={th}>{formattedProp({ value: th })}</th>)}
```
mabhub: I think `formattedProp()` should be a *(memoized)* component rather than a function.

_Anyway, not mandatory._

PR #178, opened 2020-01-06
Rearrange scene and layer fields

mabhub:


src/modules/RA/Scene/components/SceneForm.js
```diff
@@ -15,59 +17,67 @@ import {
...
 
-      <TreeInput source="tree" fullWidth />
-      {/* <TreeInput source="config.tree" fullWidth /> */}
+      <TextInput source="slug" label="view.form.slug" formClassName={classes.inline} />
+
+      <SelectInput
+        label="view.form.category"
+        source="category"
+        choices={[
+          { id: 'map', name: 'map' },
```
jrmi: Can you translate the values PLS ?

PR #179, opened 2020-01-07
Filtering viewpoints

AL9000:
Generate filters by getting filter list from the back

### DIFF
--- a/src/modules/RA/Filters/ListFilters.js
+++ b/src/modules/RA/Filters/ListFilters.js
@@ -5,34 +5,38 @@ import { fetchFilterOptions } from '../ra-modules';
 import { connectAppProvider } from '../../../components/AppProvider';
 
 
+/*
+  To generate the select input props list, first check that the filters in filterList are also
+  set in filterOptions. After that we can generate the corresponding choices list to each select
+  and generate its properties to be used on render.
+*/
+const reduceFilters = filterOptions => (filterList, [filterName, filterProps]) => {
+  const options = filterOptions[filterName];
+
+  if (!options) {
+    return filterList;
+  }
+
+  return [
+    ...filterList,
+    {
+      choices: options.map(option => ({ name: option, id: option })),
+      type: filterProps.type,
+      source: `properties__${filterProps.json_key}`,
+      label: filterProps.json_key,
+    },
+  ];
+};
+
 export const ListFilters = ({ terraOppSearchableProperties, ...props }) => {
   const [filters, setFilters] = React.useState([]);
-
-  const reduceFilters = (filterList, filterOptions) => (
-    /*
-      To generate the select input props list, first check that the filters in filterList are also
-      set in filterOptions. After that we can generate the corresponding choices list to each select
-      and generate its properties to be used on render.
-    */
-    Object.entries(filterList).reduce(
-      (reducedFilterList, [filterName, filterProps]) => {
-        if (filterName in filterOptions) {
-          reducedFilterList.push({
-            choices: filterOptions[filterName].map(value => ({ name: value, id: value })),
-            type: filterProps.type,
-            source: `properties__${filterProps.json_key}`,
-            label: filterProps.json_key,
-          });
-        }
-        return reducedFilterList;
-      }, [],
-    )
-  );
 
   React.useEffect(() => {
     const setFiltersList = async () => {
       const filterOptions = await fetchFilterOptions();
-      setFilters(reduceFilters(terraOppSearchableProperties, filterOptions));
+      const newFilters = Object.entries(terraOppSearchableProperties)
+        .reduce(reduceFilters(filterOptions), []);
+      setFilters(newFilters);
     };
 
     setFiltersList();
--- a/src/modules/RA/Filters/ListFilters.js
+++ b/src/modules/RA/Filters/ListFilters.js
@@ -2,6 +2,7 @@ import React from 'react';
 import { Filter, SelectArrayInput, SelectInput, TextInput } from 'react-admin';
 
 import { fetchFilterOptions } from '../ra-modules';
+import { connectAppProvider } from '../../../components/AppProvider';
 
 
 export const ListFilters = ({ terraOppSearchableProperties, ...props }) => {
@@ -56,4 +57,10 @@ export const ListFilters = ({ terraOppSearchableProperties, ...props }) => {
   );
 };
 
-export default ListFilters;
+const withTerraOppSearchableProperties = WrappedComponent => (
+  connectAppProvider(({ env: { terraOppSearchableProperties } }) => (
+    { terraOppSearchableProperties }
+  ))(WrappedComponent)
+);
+
+export default withTerraOppSearchableProperties(ListFilters);
--- a/src/modules/RA/Filters/ListFilters.js
+++ b/src/modules/RA/Filters/ListFilters.js
@@ -1,0 +1,59 @@
+import React from 'react';
+import { Filter, SelectArrayInput, SelectInput, TextInput } from 'react-admin';
+
+import { fetchFilterOptions } from '../ra-modules';
+
+
+export const ListFilters = ({ terraOppSearchableProperties, ...props }) => {
+  const [filters, setFilters] = React.useState([]);
+
+  const reduceFilters = (filterList, filterOptions) => (
+    /*
+      To generate the select input props list, first check that the filters in filterList are also
+      set in filterOptions. After that we can generate the corresponding choices list to each select
+      and generate its properties to be used on render.
+    */
+    Object.entries(filterList).reduce(
+      (reducedFilterList, [filterName, filterProps]) => {
+        if (filterName in filterOptions) {
+          reducedFilterList.push({
+            choices: filterOptions[filterName].map(value => ({ name: value, id: value })),
+            type: filterProps.type,
+            source: `properties__${filterProps.json_key}`,
+            label: filterProps.json_key,
+          });
+        }
+        return reducedFilterList;
+      }, [],
+    )
+  );
+
+  React.useEffect(() => {
+    const setFiltersList = async () => {
+      const filterOptions = await fetchFilterOptions();
+      setFilters(reduceFilters(terraOppSearchableProperties, filterOptions));
+    };
+
+    setFiltersList();
+  }, [terraOppSearchableProperties]);
+
+  return (
+    <Filter {...props}>
+      <TextInput label="ra.action.search" source="search" alwaysOn />
+      {filters.map(({ type, ...filterProps }) => {
+        const SelectComponent = type === 'many'
+          ? SelectArrayInput
+          : SelectInput;
+
+        return (
+          <SelectComponent
+            {...filterProps}
+            key={filterProps.label}
+          />
+        );
+      })}
+    </Filter>
+  );
+};
+
+export default ListFilters;
--- a/src/modules/RA/Filters/ListFilters.js
+++ b/src/modules/RA/Filters/ListFilters.js
@@ -32,14 +32,21 @@ export const ListFilters = ({ terraOppSearchableProperties, ...props }) => {
   const [filters, setFilters] = React.useState([]);
 
   React.useEffect(() => {
+    let isMounted = true;
+
     const setFiltersList = async () => {
       const filterOptions = await fetchFilterOptions();
+
+      if (!isMounted) return;
+
       const newFilters = Object.entries(terraOppSearchableProperties)
         .reduce(reduceFilters(filterOptions), []);
       setFilters(newFilters);
     };
 
     setFiltersList();
+
+    return () => { isMounted = false; };
   }, [terraOppSearchableProperties]);
 
   return (
--- a/src/modules/RA/Viewpoint/views/List.js
+++ b/src/modules/RA/Viewpoint/views/List.js
@@ -1,63 +1,11 @@
 import React from 'react';
-import { Filter, List, SelectArrayInput, SelectInput, TextInput } from 'react-admin';
+import { List } from 'react-admin';
 
 import CommonBulkActionButtons from '../../../../components/react-admin/CommonBulkActionButtons';
 import GridList from '../../../../components/react-admin/GridList';
-import { fetchFilterOptions } from '../../ra-modules';
 import { connectAppProvider } from '../../../../components/AppProvider';
+import ListFilters from '../../Filters/ListFilters';
 
-
-const ListFilters = ({ terraOppSearchableProperties, ...props }) => {
-  const [filters, setFilters] = React.useState([]);
-
-  const reduceFilters = (filterList, filterOptions) => (
-    /*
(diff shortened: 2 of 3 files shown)


src/modules/RA/Viewpoint/views/List.js
```diff
@@ -1,13 +1,77 @@
...
 
+
+const ListFilters = props => {
+  const [selectInputs, setSelectInputs] = React.useState([]);
+
+  const setFilters = async () => {
+    const filterOptions = await fetchFilterOptions();
+    const terraOppSearchableProperties = props.properties;
+
+    const selectInputList = Object.keys(terraOppSearchableProperties).reduce(
```
jrmi: Can you add some comment to summary the next code block ?

src/modules/RA/Viewpoint/views/List.js
```diff
@@ -1,13 +1,77 @@
...
+} from 'react-admin';
 
 import CommonBulkActionButtons from '../../../../components/react-admin/CommonBulkActionButtons';
 import GridList from '../../../../components/react-admin/GridList';
+import { fetchFilterOptions } from '../../ra-modules';
+import { connectAppProvider } from '../../../../components/AppProvider';
 
+
+const ListFilters = props => {
+  const [selectInputs, setSelectInputs] = React.useState([]);
```
jrmi: Have you considered using useMemo here instead of useState ?

src/modules/RA/ra-modules.js
```diff
@@ -119,6 +120,8 @@ export const config = {
   ],
 };
 
+export const fetchFilterOptions = async () => Api.request('viewpoints/filters/');
```
jrmi: Why do you use async here ?

src/modules/RA/Viewpoint/views/List.js
```diff
@@ -1,13 +1,77 @@
...
+  const setFilters = async () => {
+    const filterOptions = await fetchFilterOptions();
+    const terraOppSearchableProperties = props.properties;
+
+    const selectInputList = Object.keys(terraOppSearchableProperties).reduce(
+      (filters, filterName) => {
+        if (filterName in filterOptions) {
+          const choices = filterOptions[filterName].map(value => ({ name: value, id: value }));
+          if (terraOppSearchableProperties[filterName].type === 'many') {
+            filters.push(
```
mabhub: We should avoid to mix business logic and JSX. Data processing should be separated from rendering concerns.

First: Creating a variable with all requirement for rendering
Then: Render UI elements from previously created variable.

Also, rendering may be managed by a proper dedicaed component.

src/modules/RA/Viewpoint/views/List.js
```diff
@@ -1,13 +1,77 @@
...
+  return (
+    <Filter {...props}>
+      <TextInput label="ra.action.search" source="search" alwaysOn />
+      {selectInputs}
+    </Filter>
+  );
+};
+
+const ConnectedListFilters = connectAppProvider(({ env: { terraOppSearchableProperties } }) => ({
+  properties: terraOppSearchableProperties,
```
mabhub: Why renaming `terraOppSearchableProperties` to `properties` and then, renaming `properties` back to  `terraOppSearchableProperties` (L21) ?

src/modules/RA/Viewpoint/views/List.js
```diff
@@ -1,13 +1,77 @@
...
+}))(ListFilters);
 export const ViewpointList = props => (
   <List
     {...props}
     exporter={false}
+    filters={<ConnectedListFilters />}
```
mabhub:
```jsx
const terraOppSearchablePropertiesGetter = ({ env: { terraOppSearchableProperties } }) =>
  ({ terraOppSearchableProperties });

const withTerraOppSearchableProperties = WrappedComponent =>
  connectAppProvider(terraOppSearchablePropertiesGetter)(WrappedComponent);

const ListFiltersConnected = withTerraOppSearchableProperties(ListFilters);

export const ViewpointList = props => (
  <List
    {...props}
    exporter={false}
    filters={<ListFiltersConnected />}
    bulkActionButtons={<CommonBulkActionButtons />}
  >
    <GridList />
  </List>
);
```

ping @jrmi : Your thought about this proposal?

src/modules/RA/Viewpoint/views/List.js
AL9000: I will have a look to this, thanks !

src/modules/RA/Viewpoint/views/List.js
jrmi: I propose something between that stay clear but less verbose: 

```jsx
const withTerraOppSearchableProperties = WrappedComponent =>
  connectAppProvider(({ env: { terraOppSearchableProperties } }) => ({ terraOppSearchableProperties }))(WrappedComponent);

const ListFiltersConnected = withTerraOppSearchableProperties(ListFilters);

export const ViewpointList = props => (
  <List
    {...props}
    exporter={false}
    filters={<ListFiltersConnected />}
    bulkActionButtons={<CommonBulkActionButtons />}
  >
    <GridList />
  </List>
);
```

src/modules/RA/Viewpoint/views/List.js
AL9000: I just had a quick overview on the useMemo hook. Seems like an optimization to me. If so, can you tell me what is not optimized in this specific usage of the useSate hook ?

src/modules/RA/ra-modules.js
AL9000: This is a mistake.

src/modules/RA/Viewpoint/views/List.js
AL9000: This is a mistake.

src/modules/RA/Viewpoint/views/List.js
AL9000: Yes! This is in fact a necessary comment.

src/modules/RA/Viewpoint/views/List.js
```diff
@@ -1,13 +1,81 @@
...
+          />
+        ) : (
+          <SelectInput
+            source={filter.source}
+            label={filter.label}
+            choices={filter.choices}
+            key={filter.label}
+          />
+        )
+      ))}
```
mabhub: First: 
```jsx
      {filters.map(filter => {
        const SelectComponent = filter.type === 'many'
          ? SelectArrayInput
          : SelectInput;

        return (
          <SelectComponent
            source={filter.source}
            label={filter.label}
            choices={filter.choices}
            key={filter.label}
          />
        );
      })}
```

then better: 
```jsx
      {filters.map(({ type, ...filterProps }) => {
        const SelectComponent = type === 'many'
          ? SelectArrayInput
          : SelectInput;

        return (
          <SelectComponent
            {...filterProps}
            key={filterProps.label}
          />
        );
      })}
```

AL9000: :+1:

src/modules/RA/Viewpoint/views/List.js
AL9000: :+1:

src/modules/RA/Viewpoint/views/List.js
```diff
@@ -1,13 +1,80 @@
...
+
+const ListFilters = props => {
+  const [filters, setFilters] = React.useState([]);
+
+  const setFiltersList = async () => {
+    const filterOptions = await fetchFilterOptions();
+
+    // To generate the select input list to render, first check that the filters in props.properties
+    // are also set in the retrieved filter options from the backend. After that we can generate the
+    // corresponding choices list to each select and generate its properties to be used on render.
```
mabhub: Prefer block comment (`/* */`) for (pseudo)doc.
Line comment (`//`) should be kept for enabled/disabled lines.

```js
    /*
      To generate the select input list to render, first check that the filters in props.properties
      are also set in the retrieved filter options from the backend. After that we can generate the
      corresponding choices list to each select and generate its properties to be used on render. 
    */
```

AL9000: :+1:

src/modules/RA/Viewpoint/views/List.js
```diff
@@ -1,13 +1,82 @@
...
+  TextInput,
+} from 'react-admin';
 
 import CommonBulkActionButtons from '../../../../components/react-admin/CommonBulkActionButtons';
 import GridList from '../../../../components/react-admin/GridList';
+import { fetchFilterOptions } from '../../ra-modules';
+import { connectAppProvider } from '../../../../components/AppProvider';
+
+
+const ListFilters = props => {
```
mabhub:
```suggestion
const ListFilters = ({ terraOppSearchableProperties }) => {
```

src/modules/RA/Viewpoint/views/List.js
```diff
@@ -1,13 +1,82 @@
...
+
+  const setFiltersList = async () => {
+    const filterOptions = await fetchFilterOptions();
+
+    /*
+      To generate the select input list to render, first check that the filters in props.properties
+      are also set in the retrieved filter options from the backend. After that we can generate the
+      corresponding choices list to each select and generate its properties to be used on render.
+     */
+    const filterList = Object.keys(props.terraOppSearchableProperties).reduce(
```
mabhub:
```suggestion
    const filterList = Object.keys(terraOppSearchableProperties).reduce(
```

src/modules/RA/Viewpoint/views/List.js
```diff
@@ -1,13 +1,82 @@
...
+      (filtersList, filterName) => {
+        if (filterName in filterOptions) {
+          filtersList.push({
+            choices: filterOptions[filterName].map(value => ({ name: value, id: value })),
+            type: props.terraOppSearchableProperties[filterName].type,
```
mabhub:
```suggestion
            type: terraOppSearchableProperties[filterName].type,
```

src/modules/RA/Viewpoint/views/List.js
```diff
@@ -1,13 +1,82 @@
...
+            source: `properties__${props.terraOppSearchableProperties[filterName].json_key}`,
```
mabhub:
```suggestion
            source: `properties__${terraOppSearchableProperties[filterName].json_key}`,
```

src/modules/RA/Viewpoint/views/List.js
```diff
@@ -1,13 +1,82 @@
...
+            label: props.terraOppSearchableProperties[filterName].json_key,
```
mabhub:
```suggestion
            label: terraOppSearchableProperties[filterName].json_key,
```

src/modules/RA/Filters/ListFilters.js
```diff
@@ -0,0 +1,59 @@
...
+        return (
+          <SelectComponent
+            {...filterProps}
+            key={filterProps.label}
+          />
+        );
+      })}
+    </Filter>
+  );
+};
```
mabhub: Take a look at: 

```jsx
import React from 'react';
import { Filter, SelectArrayInput, SelectInput, TextInput } from 'react-admin';

import { fetchFilterOptions } from '../ra-modules';

/*
  To generate the select input props list, first check that the filters in filterList are also
  set in filterOptions. After that we can generate the corresponding choices list to each select
  and generate its properties to be used on render.
*/
const reduceFilters = filterOptions => (reducedFilterList, [filterName, filterProps]) => {
  if (filterName in filterOptions) {
    reducedFilterList.push({
      choices: filterOptions[filterName].map(value => ({ name: value, id: value })),
      type: filterProps.type,
      source: `properties__${filterProps.json_key}`,
      label: filterProps.json_key,
    });
  }
  return reducedFilterList;
};

export const ListFilters = ({ terraOppSearchableProperties, ...props }) => {
  const [filters, setFilters] = React.useState([]);

  React.useEffect(() => {
    const setFiltersList = async () => {
      const filterOptions = await fetchFilterOptions();
      const newFilters = Object.entries(terraOppSearchableProperties)
        .reduce(reduceFilters(filterOptions), []);
      setFilters(newFilters);
    };

    setFiltersList();
  }, [terraOppSearchableProperties]);

  return (
    <Filter {...props}>
      <TextInput label="ra.action.search" source="search" alwaysOn />
      {filters.map(({ type, ...filterProps }) => {
        const SelectComponent = type === 'many'
          ? SelectArrayInput
          : SelectInput;

        return (
          <SelectComponent
            {...filterProps}
            key={filterProps.label}
          />
        );
      })}
    </Filter>
  );
};

export default ListFilters;
```

mabhub: Also, `reduceFilters` may be written lie this:

```jsx
const reduceFilters = filterOptions => (filterList, [filterName, filterProps]) => {
  const options = filterOptions[filterName];

  if (!options) {
    return filterList;
  }

  return [
    ...filterList,
    {
      choices: options.map(option => ({ name: option, id: option })),
      type: filterProps.type,
      source: `properties__${filterProps.json_key}`,
      label: filterProps.json_key,
    },
  ];
};
```

src/modules/RA/Viewpoint/views/List.js
jrmi: It's not "just optimization", I think it can completely replace useState here. The state is only used to save the last value, which seems to be the exact use case of "useMemo". You can read [this](https://wattenberger.com/blog/react-hooks) for more details.

jrmi: Discussed with @mabhub, i didn't see the async call. So there is an other problem : the setState can be called on an unmounted component so you need to apply the pattern described here: https://stackoverflow.com/a/56747072

src/modules/RA/Filters/ListFilters.js
```diff
@@ -0,0 +1,70 @@
...
+  React.useEffect(() => {
+    const setFiltersList = async () => {
+      const filterOptions = await fetchFilterOptions();
+      const newFilters = Object.entries(terraOppSearchableProperties)
+        .reduce(reduceFilters(filterOptions), []);
+      setFilters(newFilters);
+    };
+
+    setFiltersList();
+  }, [terraOppSearchableProperties]);
```
jrmi:
```
 React.useEffect(() => {
    let isMounted = true;
    const setFiltersList = async () => {
      const filterOptions = await fetchFilterOptions();
      if (!isMounted) return;
      const newFilters = Object.entries(terraOppSearchableProperties)
        .reduce(reduceFilters(filterOptions), []);
      setFilters(newFilters);
    };

    setFiltersList();
    return () => {
        isMounted = false;
    }
  }, [terraOppSearchableProperties]);
```

src/modules/RA/Viewpoint/views/List.js
jrmi: I wrote another comment with the modification.

src/modules/RA/Filters/ListFilters.js
AL9000: Done, thx :)

PR #181, opened 2020-01-10
Set layerStyle defaultvalue on create

dtrucs:


src/modules/RA/DataLayer/components/DataLayerTabbedForm.js
```diff
@@ -50,10 +52,21 @@ const CustomFormTab = ({ hidden, ...props }) => (
     : <FormTab hidden={hidden} {...props} />
 );
 
-const DataLayerTabbedForm = ({ classes, translate, ...props }) => (
+const DataLayerTabbedForm = ({
+  classes,
+  translate,
+  randomColor,
+  sourceData: { geom_type: type = POLYGON } = {},
```
mabhub:
```suggestion
  sourceData: { geom_type: geomType = POLYGON } = {},
```

src/modules/RA/DataLayer/components/DataLayerTabbedForm.js
```diff
@@ -50,10 +52,21 @@ const CustomFormTab = ({ hidden, ...props }) => (
...
+  ...props
+}) => (
   <>
     <SourceFetcher />
-    <TabbedForm {...props}>
+    <TabbedForm
+      defaultValue={{
+        layer_style: getLayerStyleDefaultValue(randomColor, type),
```
mabhub:
```suggestion
        layer_style: getLayerStyleDefaultValue(randomColor, geomType),
```

src/utils/geom.js
```diff
@@ -7,6 +7,29 @@ export const MULTI_LINESTRING = 5;
...
+export const getLayerStyleDefaultValue = (color, geomType) => {
+  const type = getShapeFromGeomType(geomType);
+  return {
+    type,
+    paint: {
+      [`${type}-color`]: color,
+    },
+  };
+};
+
```
mabhub: Just a suggestion, not mandatory:

```jsx
export const getLayerStyleDefaultValue_ = (
  color,
  geomType,
  type = getShapeFromGeomType(geomType),
) => ({
  type,
  paint: {
    [`${type}-color`]: color,
  },
});
```

dtrucs: I'm not sure about this one.
The `geomType` argument is only useful for getting the shape, we don't need it for anything else.
If we set the third parameters `type`, the `geomType` in the middle seems confusing.

mabhub: Could also be done with object as function arg:

```jsx
export const getLayerStyleDefaultValue = ({
  color,
  geomType,
  type = getShapeFromGeomType(geomType),
}) => ({
  type,
  paint: {
    [`${type}-color`]: color,
  },
});
```

PR #183, opened 2020-01-16
Improve datatable header component

dtrucs:


src/modules/CRUD/components/DataTable/Header/index.js
```diff
@@ -20,5 +21,6 @@ export default compose(
   withRouter,
   connectAuthProvider(authProviderGetter),
   connectCRUDProvider('featuresList'),
+  connectUserSettingsProvider('tableSize', 'setTableSize'),
```
mabhub: This could be replaced by a simple function:

```suggestion
  withTableSize(),
```

*not mandatory.*

src/modules/CRUD/services/UserSettingsProvider.js
```diff
@@ -11,10 +11,13 @@ const { Provider } = context;
 
 export const UserSettingsProvider = ({ children }) => {
   const [pageSize, setPageSize] = useState(10);
+  const [tableSize, setTableSize] = useState('medium'); // 'minified', 'medium', 'full'
```
mabhub: Couldn't these string values be replaced by constant equivalent?

```js
export const MINIFIED = 'minified';
export const MEDIUM = 'medium';
export const FULL = 'full';
```

PR #186, opened 2020-01-20
Fix broken filter config form

jrmi:


src/components/react-admin/DraggableFormIterator.js
```diff
@@ -33,10 +32,14 @@ import { CSSTransition, TransitionGroup } from 'react-transition-group';
 import compose from '../../utils/compose';
 import './draggable.scss';
 
-const DragHandle = sortableHandle(() => <DragHandleIcon />);
+const DragHandle = sortableHandle(({ classes }) => (
+  <span className={classes.dragHandle}>
+    <DragHandleIcon />
+  </span>
+));
```
dtrucs: Maybe, ou can render this component dumber :
```
const DragHandle = sortableHandle(props => (
  <span {...props}>
    <DragHandleIcon />
  </span>
));
```

And its call below
```
<DragHandle className={classes.dragHandle} />
```

src/components/react-admin/DraggableFormIterator.js
```diff
@@ -205,9 +213,7 @@ export class DraggableFormIterator extends Component {
     // the fields prop which will always be empty for a new record.
     // Without it, our ids wouldn't match the default value and we would get key warnings
     // on the CssTransition element inside our render method
-    this.ids = this.nextId > 0
-      ? Array.from(Array(this.nextId).keys())
-      : [];
+    this.ids = this.nextId > 0 ? Array.from(Array(this.nextId).keys()) : [];
```
dtrucs: For the same goal, maybe this is more readable to understand the meaning:
`this.ids = this.nextId > 0 ? Array.from({ length: this.nextId }, (_, index) => index) : [];`

jrmi: I think I can also remove the "this.nextId>0" as it's safe to call this with `0` value.

I propose some shorter alternatives. Which one do you prefer ?

```js
// 1 Use spreading operator
this.ids = [...Array(5).keys()];

// 2 Use lodash range
import range from 'lodash'
this.ids = range(5);

File: src/components/react-admin/DraggableFormIterator.js (hunk 1 above)
Comment (jrmi) [reply src/components/react-admin/DraggableFormIterator.js]: done

File: src/components/react-admin/DraggableFormIterator.js (hunk 2 above)
Comment (dtrucs) [reply src/components/react-admin/DraggableFormIterator.js]: I also think it's safe but i don't really know how ids are defined so i though it's not bad to keep `> 0`. Moreover, it gives the information that `this.nextId` is an integer type.

About 1/, i prefer my request, but it's equal.
About 2/, as we said IRL, the use of lodash is a little overkill.

File: src/modules/RA/DataLayer/components/SourceFetcher.js [hunk 3]
```diff
@@ -19,10 +19,14 @@ const SourceFetcher = ({ dispatch, dataProvider, sourceId, fields = [] }) => {
   const load = memo(async id => dataProvider(GET_ONE, RES_DATASOURCE, { id }));
 
   useEffect(() => {
-    if (!sourceId) return;
+    if (!sourceId) return () => {};
```
mabhub:
```suggestion
    if (!sourceId) return null;
```

PR #188, opened 2020-01-22
Fix main field and layer fields from layer

jrmi:


src/modules/RA/DataLayer/components/DataLayerTabbedForm.js
```diff
@@ -100,7 +100,7 @@ const DataLayerTabbedForm = ({
                 <SelectInput
                   source="main_field"
                   label="datalayer.form.search.main-field.label"
-                  choices={formData.fields.map(({ label: name, id }) => ({ id, name }))}
+                  choices={formData.fields.map(({ label: name, field }) => ({ id: field, name }))}
```
dtrucs: Try to be consistent on spreading:
`formData.fields.map(({ field: id, label: name}) => ({ id, name }))}`
or 
`formData.fields.map(({ field, label }) => ({ id: field, name: label }))}`

jrmi: You are so damn right. Fixed.

PR #191, opened 2020-01-29
Search in data table

dtrucs:


src/modules/CRUD/components/DataTable/Footer/index.js
```diff
@@ -1,9 +1,9 @@
 import { withNamespaces } from 'react-i18next';
-import { withPageSize } from '../../../services/UserSettingsProvider';
+import { connectUserSettingsProvider } from '../../../services/UserSettingsProvider';
 import compose from '../../../../../utils/compose';
 import Footer from './Footer';
 
 export default compose(
-  withPageSize(),
+  connectUserSettingsProvider('tableFilters', 'setTableFilters', 'pageSize', 'setPageSize'),
```
mabhub: Why not using a `withTableFilter()` _hoc_ like `withPageSize()` ?

src/modules/CRUD/components/DataTable/Search/Search.js
```diff
@@ -0,0 +1,85 @@
...
+import PropTypes from 'prop-types';
+import debounce from 'lodash.debounce';
+import { Icon, InputGroup, Spinner } from '@blueprintjs/core';
+import './styles.scss';
+
+const isSearchingOrNot = isSearching => (
+  isSearching
+    ? <Spinner size={Icon.SIZE_STANDARD} />
+    : <div style={{ width: 30 }} />
+);
```
mabhub:
```jsx
const CustomLoader = ({ loading, ...props }) => (
  loading
    ? <Spinner size={Icon.SIZE_STANDARD} {...props} />
    : <div style={{ width: 30 }} {...props} />
); 
```

src/modules/CRUD/components/DataTable/Search/Search.js
```diff
@@ -0,0 +1,85 @@
...
+  };
+
+  return (
+    <InputGroup
+      defaultValue={defaultValue}
+      inputRef={inputRef}
+      onChange={onChange}
+      minimal
+      leftIcon="search"
+      rightElement={isSearchingOrNot(isSearching)}
```
mabhub:
```suggestion
      rightElement={<CustomLoader loading={isSearching} />}
```

src/modules/CRUD/components/DataTable/Footer/index.js
dtrucs: It may be better for readability purposes, but I think that calling the same HOC twice is a bit of overkill, isn't it?

src/modules/CRUD/services/UserSettingsProvider.js
```diff
@@ -16,12 +16,15 @@ const { Provider } = context;
 export const UserSettingsProvider = ({ children }) => {
   const [pageSize, setPageSize] = useState(10);
   const [tableSize, setTableSize] = useState(TABLE_MEDIUM);
+  const [tableFilters, setTableFilters] = useState({});
```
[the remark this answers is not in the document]
dtrucs: Ok I'll try

src/modules/CRUD/components/DataTable/Footer/index.js
mabhub: I didn't do any test about that point, but yes, it's a point. I don't have any idea if setting multiple times the same HOC may be an issue or not. It should be checked...

But I think that the real pattern that we should use in this case is custom hooks *(`const [pageSize, setPageSize] = usePageSize();`)* instead of HOCs. But we do not have define any guideline about that for now.

So, I would suggest you to quickly test *if* setting the same HOCs seems implying perf/rendering issues.
If not, `withThisData()` is clearly best for DX and maintainability. If any suspicion of negative impact, keep your initial choice.

PR #193, opened 2020-01-31
Fix default style again

jrmi:
With last modification, default style was used even if the style is defined.

### DIFF
--- a/src/modules/RA/DataLayer/components/DataLayerTabbedForm.js
+++ b/src/modules/RA/DataLayer/components/DataLayerTabbedForm.js
@@ -113,14 +113,13 @@ const DataLayerTabbedForm = ({
 
         <FormDataConsumer>
           {({ formData }) =>
+            // Allow to initialize default value even if next tab is not yet loaded
             formData.source && (
             <TextInput
               type="hidden"
               label=""
               source="layer_style"
-              defaultValue={() => ({
-                layer_style: getLayerStyleDefaultValue(randomColor, getShapeFromGeomType(geomType)),
-              })}
+              defaultValue={getLayerStyleDefaultValue(randomColor, getShapeFromGeomType(geomType))}
             />
             )
           }
@@ -134,6 +133,7 @@ const DataLayerTabbedForm = ({
           source="layer_style"
           withSource="source"
           label="datalayer.form.styles.mainstyle"
+          defaultValue={getLayerStyleDefaultValue(randomColor, getShapeFromGeomType(geomType))}
           fullWidth
         />
 
--- a/src/modules/RA/DataLayer/components/DataLayerTabbedForm.js
+++ b/src/modules/RA/DataLayer/components/DataLayerTabbedForm.js
@@ -48,9 +48,7 @@ const styles = {
 
 // Best performance when tabs have heavy content and produce multiple render
 const LazyFormTab = ({ hidden, ...props }) => (
-  hidden
-    ? null
-    : <FormTab hidden={hidden} {...props} />
+  hidden ? null : <FormTab {...props} />
 );
 
 const DataLayerTabbedForm = ({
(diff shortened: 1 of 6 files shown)


src/modules/RA/DataLayer/components/DataLayerTabbedForm.js
```diff
@@ -46,7 +46,8 @@ const styles = {
   },
 };
 
-const CustomFormTab = ({ hidden, ...props }) => (
+// Best performance when tabs have heavy content and produce multiple render
+const LazyFormTab = ({ hidden, ...props }) => (
   hidden
     ? null
     : <FormTab hidden={hidden} {...props} />
```
dtrucs: If hidden is a boolean value, you can one line the return. And the hidden prop would be useless
`hidden && <FormTab {...props} />`

src/modules/RA/DataLayer/components/DataLayerTabbedForm.js
```diff
@@ -111,13 +110,32 @@ const DataLayerTabbedForm = ({
...
+
+      </LazyFormTab>
+
+      <LazyFormTab label="datalayer.form.style" path="style">
+
         <StyleField
           source="layer_style"
           withSource="source"
           label="datalayer.form.styles.mainstyle"
+          defaultValue={getLayerStyleDefaultValue(randomColor, getShapeFromGeomType(geomType))}
```
dtrucs: Maybe you can declare the default value and call it in this line and line 123
`const layerStyleDefaultValue = getLayerStyleDefaultValue(randomColor, getShapeFromGeomType(geomType))`

src/modules/RA/DataLayer/components/StyleField.js
```diff
@@ -9,9 +9,6 @@ const validate = value => {
   if (!value.type) {
     return 'type not found';
   }
-  if (!value.paint) {
-    return 'paint not found';
-  }
```
dtrucs: Sure of this?

dtrucs: Also, maybe some translations would be appropriate.

src/modules/RA/DataLayer/components/DataLayerTabbedForm.js
dtrucs: Finally, it wouldn't work for the first suggestion, but it could be still relevant for the secnd part:

```jsx
const LazyFormTab = ({ hidden, ...props }) => (
  hidden
    ? null
    : <FormTab {...props} />
```

jrmi: I get this error with your code : `Invariant Violation: LazyFormTab(...): Nothing was returned from render`

src/modules/RA/DataLayer/components/DataLayerTabbedForm.js
jrmi: There is side effect here as the source is set later so the default value is invalid. Here the value is set only when the tab is opened.

src/modules/RA/DataLayer/components/StyleField.js
jrmi: Yes sure, we have case in PACA where only icon are added and paint is empty. But you are right for translation.

src/modules/RA/DataLayer/components/DataLayerTabbedForm.js
jrmi: Ok for the second part !

src/modules/RA/DataLayer/components/StyleField.js
jrmi: done

src/modules/RA/DataLayer/components/DataLayerTabbedForm.js
jrmi: Done

PR #196, opened 2020-02-04
Store user settings

dtrucs:


src/modules/CRUD/services/UserSettingsProvider.js
```diff
@@ -8,20 +8,27 @@ export const TABLE_MINIFIED = 'minified';
...
-  const [pageSize, setPageSize] = useState(10);
-  const [tableSize, setTableSize] = useState(TABLE_MEDIUM);
-  const [tableFilters, setTableFilters] = useState({});
+  const [tableFilters, setTableFilters] = useState(getUserSettings('tableFilters') || {
+    page_size: 10,
+  });
+  const [tableSize, setTableSize] = useState(getUserSettings('tableSize') || TABLE_MEDIUM);
+
+  useEffect(() => {
+    localStorage.setItem('CRUDUserSettings', JSON.stringify({
```
mabhub: I'm not really sure that using a provider pattern for data stored in localStorage is really a good way to do it...

dtrucs: The final goal is to get and store these settings via API calls. It's a sort of patch

PR #197, opened 2020-02-05
Fix layer fields initialisation from geosource fields

jrmi:
Was broken when creating a new source.


src/modules/RA/DataLayer/components/DataLayerTabbedForm.js
```diff
@@ -56,25 +58,33 @@ const DataLayerTabbedForm = ({
...
-          label="datalayer.form.data-source"
-          sort={{ field: 'name', order: 'ASC' }}
-          validate={defaultRequired}
-          perPage={100}
-        >
-          <SelectInput />
-        </ReferenceInput>
+      <LazyFormTab
+        label="datalayer.form.definition"
+      >
```
mabhub: Could we let this one inline?
```jsx
      <LazyFormTab label="datalayer.form.definition"> 
```

src/modules/RA/DataLayer/components/FieldUpdater.js
```diff
@@ -0,0 +1,115 @@
...
+};
+
+export const updateFieldFromSource = async (layerFields, dispatch, dataProvider, sourceId) => {
+  const { data: { fields: sourceFields = [] } } =
+      await dataProvider(GET_ONE, RES_DATASOURCE, { id: sourceId });
+
+  const sourceFieldMapped = sourceFields.map(({ id, ...sourceField }) =>
+    ({ field: id, sourceFieldId: id, ...sourceField }));
+
+  const result = mergeFields(layerFields || [], sourceFieldMapped);
```
dtrucs: To avoid this kind of trick `|| []`, maybe you can initialise default value in the function line 25 ⬇️ 
`const mergeFields = (layerFields = [], sourceFields = []) => {`

dtrucs: And so, in this function you can make an early return
`if(!layerFields.length && !sourceFields.length) return []`

jrmi: Done for first comment but for the second comment this case is so rare (should never happens) that i don't want an useless test.

PR #206, opened 2020-03-05
Upgrade packages and forward locale to MapBox controls

dtrucs:


mabhub:
Couldn't we also upgrade `react-i18next`? _(So we'd have translation hooks available!)_

dtrucs:
> Couldn't we also upgrade `react-i18next`? _(So we'd have translation hooks available!)_

Due to a large rewriting to be compatible with the last version of `react-i18next`, @mabhub  could you test if everything is fine in RA side?

PR #218, opened 2020-03-23
Refacto create feature

dtrucs:


src/components/react-json-schemaForm/GeometryField/PointField/PointField.js
```diff
@@ -10,46 +10,62 @@ const Label = ({ name, required }) => (
...
   schema: {
     required = [],
   },
   disabled,
 }) => {
-  const [coords, setCoords] = useState(formData.coordinates);
   const isRequired = required.includes('coordinates');
 
+  const refLatitude = useRef(null);
+  const refLongitude = useRef(null);
```
mabhub: Why using useRefs for numbers/strings?
Objects & Arrays may need refs because shallow comparision (`===`) is always false. But numbers and strings do not have this issue.

src/components/react-json-schemaForm/GeometryField/PointField/PointField.js
```diff
@@ -10,46 +10,62 @@ const Label = ({ name, required }) => (
...
-  const handleChange = index => ({ target: { value } }) => {
-    const coordinates = [...coords];
-    coordinates[index] = Number(value);
-    setCoords(coordinates);
+  useEffect(() => {
+    const [latitude = '', longitude = ''] = coordinates;
+    refLatitude.current.value = latitude;
+    refLongitude.current.value = longitude;
+    onChange(formData);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
jrmi: Again, why do you ignore onChange change ?

src/components/react-json-schemaForm/GeometryField/PointField/PointField.js
```diff
@@ -10,46 +10,62 @@ const Label = ({ name, required }) => (
...
+  }, [coordinates]);
+
+
+  const handleChange = () => {
```
jrmi: You must use `useCallback` here.

src/modules/CRUD/components/Details/Create/Create.js
```diff
@@ -0,0 +1,186 @@
...
+  const isMounted = useRef(true);
+
+  useEffect(() => {
+    isMounted.current = true;
+    return () => { isMounted.current = false; };
+  }, []);
+
+  const title = t('CRUD.details.create', { layer: name });
+
+  const [loading, setLoading] = useState(false);
```
jrmi: I think hooks should be at the top of component, just after or before the `useRef`.

src/modules/CRUD/components/Details/Create/Create.js
```diff
@@ -0,0 +1,186 @@
...
+      <Header title={title} />
+      <div className="details__list-create">
+        <Form
+          className="CRUD-edit CRUD-edit--vertical"
+          disabled={loading}
+          schema={schema}
+          uiSchema={uiSchema}
+          fields={{ rte: RTEField, geometry: GeometryField }}
+          onSubmit={
+            submitFeature({ isMounted, setLoading, setSchema, ...props })
```
jrmi: I think you also need "useCallback" here.

src/components/react-json-schemaForm/GeometryField/GeometryField.js
```diff
@@ -1,45 +1,208 @@
...
 
-const Fieldset = ({ t, children }) => (
-  <fieldset>
-    <legend>{t('jsonSchema.geometryField.title')}</legend>
-    <div className="form-group field">
-      {children}
-    </div>
-  </fieldset>
-);
+const updateGeometry = ({
```
jrmi: Can you add some comments ?

src/components/react-json-schemaForm/GeometryField/GeometryField.js
```diff
@@ -1,45 +1,208 @@
...
   }
-  const action = !coordinates.length ? 'create' : 'edit';
+
+  setGeomValues(prevGeomValues => ({
+    ...prevGeomValues,
+    geom: geometry,
+  }));
+};
+
+const initDraw = ({
```
jrmi: Can you add comments ?

src/components/react-json-schemaForm/GeometryField/GeometryField.js
```diff
@@ -1,45 +1,208 @@
...
+
+const resetStyle = (map, layerPaints, name) => {
+  if (!Object.keys(map).length) {
+    return;
+  }
+  const layerPaintId = getLayerPaintId(layerPaints, name);
+  map.setFilter(layerPaintId, null);
+};
+
+const getGeometries = ({
```
jrmi: Can you add comments ?

src/components/react-json-schemaForm/GeometryField/PointField/PointField.js
```diff
@@ -10,46 +10,62 @@ const Label = ({ name, required }) => (
...
+              inputMode="numeric"
+              inputRef={ref}
+              onChange={handleChange}
+              pattern="[+-]?([0-9]*[.])?[0-9]+"
+              required={isRequired}
+              type="text"
+            />
+          </FormGroup>
+        );
+      })}
```
mabhub: Using a ref for each fields seems to be here to allow the `[].map()`.
You could avoid this quite useless ref by just using a simpler intermediate component :

```jsx
const CustomCoordField = ({ name, id, ...props }) => {
  const { required } = props;

  return (
          <FormGroup
            label={<Label name={t(`jsonSchema.geometryField.${name}`)} required={required} />}
            labelFor={`${$id}-${name}`}
          >
            <InputGroup
              id={`${$id}-${name}`}
              inputMode="numeric"
              pattern="[+-]?([0-9]*[.])?[0-9]+"
              type="text"
              {...props}
            />
          </FormGroup>
  );
};

// ...

const PointField = ({
  // ...
}) => {
  // ...
  return (
    <div>
      <CustomCoordField
        name="latitude"
        onChange={handleChange}
        required={isRequired}
        defaultValue={coordinates[0]}
      />
      <CustomCoordField
        name="longitude"
        onChange={handleChange}
        required={isRequired}
        defaultValue={coordinates[1]}
      />
    </div>
  );
};
```

src/modules/CRUD/components/Details/Create/Create.js
```diff
@@ -0,0 +1,186 @@
...
+import Form from 'react-jsonschema-form';
+import { getJSONSchemaFromGeom } from '../../../services/utils';
+import { toast } from '../../../../../utils/toast';
+import { generateURI } from '../../../config';
+import RTEField from '../../../../../components/react-json-schemaForm/RTEField';
+import GeometryField from '../../../../../components/react-json-schemaForm/GeometryField';
+import Header from '../Header';
+
+
+const submitFeature = props => async ({ formData: { geom, ...properties } }) => {
```
jrmi: Can you add comment here ?

src/modules/CRUD/components/Details/Create/Create.js
```diff
@@ -0,0 +1,186 @@
...
+  if (savedFeature) {
+    push(generateURI('layer', { layer: name, id: savedFeature.identifier }));
+    toast.displayToaster(
+      { id: savedFeature.identifier },
+      t('CRUD.details.successCreateFeature'),
+    );
+  }
+};
+
+const buildProperties = props => {
```
jrmi: Can you add comments to all big functions PLS ?

src/components/react-json-schemaForm/GeometryField/DefaultField/DefaultField.js
```diff
@@ -0,0 +1,51 @@
...
+    formData,
+    formData: { coordinates, type } = {},
+    onChange,
+    t,
+  } = props;
+
+  useEffect(() => {
+    onChange(formData);
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [formData]);
```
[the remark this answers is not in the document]
dtrucs: With onChange it renders an infinite loop, the workaround i found is to useCallback the onChange prop.
It seems hacky, maybe you know another tip?

src/components/react-json-schemaForm/GeometryField/PointField/PointField.js
dtrucs: This is not the case.
If the user draw the point in the map, i use ref to set coordinates on each field (cf the useEffect at the top)

```js
  useEffect(() => {
    const [latitude = '', longitude = ''] = coordinates;
    refLatitude.current.value = latitude;
    refLongitude.current.value = longitude;
    onChange(formData);
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, [coordinates]);
```

dtrucs: But i check to refacto this

src/components/react-json-schemaForm/GeometryField/DefaultField/DefaultField.js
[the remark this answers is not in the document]
dtrucs: Finally, it doesn't render in an infinite way. 
But each time the useEffect is called twice with `onChange` in the second's array parameter (versus once without it).

src/components/react-json-schemaForm/GeometryField/GeometryField.js
```diff
@@ -1,48 +1,219 @@
...
+      params,
+      name,
+      settings,
+    }))
+  ), [feature, formData, params, name, settings]);
+
+  const [geomValues, setGeomValues] = useState(geometries);
+
+  useEffect(() => {
+    setGeomValues(geometries);
```
jrmi: But the default value line 94 isn't enougth here ?

dtrucs: I want to update this state only when url (`params`) is changing

jrmi: Okay.

src/components/react-json-schemaForm/GeometryField/GeometryField.js
```diff
@@ -1,48 +1,219 @@
...
+    formData,
+    layerPaints,
+    map,
+    match: { params },
+    removeControl,
+    name,
+    settings,
+    t,
+    ...rest
+  } = props;
```
mabhub:
```suggestion
const GeometryField = ({
  addControl,
  feature,
  formData,
  layerPaints,
  map,
  match: { params },
  removeControl,
  name,
  settings,
  t,
  ...rest
}) => {
```

src/components/react-json-schemaForm/GeometryField/GeometryField.js
```diff
@@ -1,48 +1,219 @@
...
+
+  const geometries = useMemo(() => (
+    getGeometries(({
+      feature,
+      formData,
+      params,
+      name,
+      settings,
+    }))
+  ), [feature, formData, params, name, settings]);
```
mabhub: Inlining object properties could improve visual similarity with hook dependencies.

```suggestion
  const geometries = useMemo(() => (
    getGeometries(({ feature, formData, params, name, settings }))
  ), [feature, formData, params, name, settings]);
```

src/components/react-json-schemaForm/GeometryField/GeometryField.js
```diff
@@ -1,48 +1,219 @@
...
+    const { features } = target.draw.getAll();
+
+    if (features.length > 1 && [POINT, LINESTRING, POLYGON].includes(geomValues.geom_type)) {
+      target.draw.delete(
+        features.reduce((list, item) => (
+          item.id !== id
+            ? [...list, item.id]
+            : list
+        ), []),
+      );
```
mabhub: Doing this in two steps improve understandability.
*(also, simpler case first)*

```suggestion
      const otherFeaturesIds = features.reduce((list, item) => (
        item.id === id ? list : [...list, item.id]
      ), []);
        
      target.draw.delete(otherFeaturesIds);
```

src/components/react-json-schemaForm/GeometryField/PointField/PointField.js
mabhub: I still do not see the point of using refs this way: if you have to manage the value of inputs, using value prop of each one and corresponding states is the classical way.
And if you have multiple states having to be changed together, `useReducer()` is the best solution.
I probably miss something else...

src/components/react-json-schemaForm/GeometryField/index.js
```diff
@@ -1,4 +1,26 @@
...
+  feature,
+  map,
+  settings,
+}) =>
+  ({
+    feature,
+    layerPaints: getLayersPaints(settings),
+    map,
+    settings,
+  });
```
mabhub: I think this would be clearer:

```suggestion
const CRUDPRoviderGetter = ({ feature, map, settings }) =>
  ({ feature, map, settings, layerPaints: getLayersPaints(settings) });

```

src/modules/CRUD/components/Details/Create/Create.js
```diff
@@ -0,0 +1,194 @@
...
+    setSchema(prevSchema => {
+      const defaultValues = Object.entries(properties);
+      if (defaultValues.length) {
+        const nextSchema = { ...prevSchema };
+        defaultValues.forEach(([key, value]) => {
+          nextSchema.properties[key].default = value;
+        });
+        return nextSchema;
+      }
+      return prevSchema;
```
mabhub: This could be a dedicated (pure) function with `properties` and `prevSchema` as arguments. The name of this function would then *explain* what transformation is done.

dtrucs: Right, done.

PR #219, opened 2020-03-31
Create table fields component

dtrucs:


src/components/react-json-schemaForm/TableField/TableField.js
```diff
@@ -0,0 +1,69 @@
...
+            {...props}
+            formData={data}
+            onChange={handleChange}
+            onUpdatingRow={setData}
+          />
+        </table>
+      </div>
+    </fieldset>
+  );
+};
```
mabhub: The reason why *original* `props` are used for `<Thead />` but  *altered* props in `<Tbody />` is not clearly understandable. I don't say this has to be changed or is wrong. But it adds some unecessary complexity.

I would maybe have done a first wrapping (dump) stateless component with `<fieldset />`, `title`, `<table />` and `<Thead {...props} />`, then, another component holding state and `handleChange` method, and called it in place of `<Tbody />`.


```jsx
const CustomTbody = ({ formData, onChange, ...props }) => {
  const [data, setData] = useState(formData);

  useEffect(() => {
    onChange(data);
  }, [data, onChange]);

  const handleChange = useCallback((row, key) => value => {
    const nextData = [...data];
    nextData[row][key] = value;
    onChange(nextData);
  }, [data, onChange]);

  return (
    <Tbody
      formData={data}
      onChange={handleChange}
      onUpdatingRow={setData}
      {...props}
    />
  );
};


const TableField = props => (
  <fieldset className="field field-array field-array-of-object">
    <legend>{title}</legend>
    <div className="details__table tableField">
      <table>
        <Thead {...props} />
        <CustomTbody {...props} />
      </table>
    </div>
  </fieldset>
);
```

src/components/react-json-schemaForm/TableField/Tbody/Tbody.js
```diff
@@ -0,0 +1,150 @@
...
+
+
+  if (!data.length) {
+    handleUpdatingRow(0, 'inserting');
+  }
+
+  return (
+    <tbody>
+      {data.map((item, index) => {
+        const orderedValues = getObjectOrderedValue(properties, uiSchemaItems['ui:order']);
```
mabhub: Why is this computed for *each* iteration of `data.map` when it does not use any loop data?

dtrucs: Right, this is a mistake.
Moreover, it's also declared line 36.

I'll refacto and memoize the constante at the top of the component 👍

src/components/react-json-schemaForm/TableField/TableField.js
dtrucs: Right, 
Here, there was some logic which shouldn't be.
 i just moved the `useEffect` and the `handlechange` directly in the `Tbody` component

src/components/react-json-schemaForm/TableField/TableField.js
```diff
@@ -0,0 +1,49 @@
...
+TableField.defaultProps = {
+  schema: {
+    items: {
+      properties: {},
+      required: [],
+    },
+    title: '',
+  },
+  onChange: () => {},
+};
```
mabhub: Shouldn't those props be imported from `<Thead />` or `<Tbody />` component instead of being duplicated here?

dtrucs: Understood,
How do you suggest the propType/defaultProp of `title` ? Something like ⬇️ 
```
const allDefaultProps = { ...Tbody.defaultProps, ...Thead.defaultProps }
TableField.defaultProps = {
  ...allDefaultProps,
  schema: {
    ...allDefaultProps.schema,
    title: ''
  }
};
```
Maybe, it lose some readability

mabhub: I think I would personaly not bother with propTypes, and only provide default value directly when destructuring:
```jsx
const TableField = props => {
  const {
    schema: { title = '' },
  } = props;

  // ...
```

*(I also think that `schema` and `uiSchema` propTypes shouldn't be defined each type a Component use them. They should be defined only once, then import and used when needed. Also, tools for generating propTypes from JSONSchema exist)*

dtrucs: Ok done,
Finally, i didn't set a default prop to title in benefit of a conditional display of this title

src/components/react-json-schemaForm/TableField/Field/Field.js
```diff
@@ -0,0 +1,54 @@
...
+    formData,
+    onChange,
+    registry: { fields: { SchemaField } },
+  } = props;
+
+  const [data, setData] = useState(formData);
+
+  const handleChange = value => {
+    setData(value);
+  };
```
jrmi: You need useCallback here 
```suggestion
  const handleChange = useCallback(value => {
    setData(value);
  }, []);
```

or you can also just keep the setData: 
```suggestion
  const handleChange = setData;
```
which is also working.

dtrucs: Good point ! 👍 
I chose the third option ⬇️ 
```js
<SchemaField onChange={setData} />
```

PR #223, opened 2020-04-02
Improve tbody component

dtrucs:


src/components/react-json-schemaForm/TableField/Tbody/Tbody.js
```diff
@@ -1,7 +1,7 @@
 import React, { useState, useEffect, useCallback, useMemo } from 'react';
 import PropTypes from 'prop-types';
 import { Button } from '@blueprintjs/core';
-import uuid from 'uuid/v4';
+import { v4 as uuid } from 'uuid';
```
mabhub, line 4: Does this change really make sense?

*Generally*, `import { z as x } from 'y';` imply including the whole `y` source in JavaScript bundle, 
when `import x from 'y/z';` *(when made available by package author)* will only include the required part of source, instead of full package.

dtrucs: I dunno but before this change, i have the warning in the console
>Deep requiring like `const uuidv4 = require('uuid/v4');` is deprecated as of uuid@7.x. Please require the top-level module when using the Node.js CommonJS module or use ECMAScript Modules when bundling for the browser. See https://github.com/uuidjs/uuid#deep-requires-now-deprecated for more information.

mabhub: The link you gave in your quote explain exactly that package author by using ECMAScript module (instead of CommonJS module, which is still more *common*), allow JS bundler to *tree-shake* also when importing top level module (which is not possible with CommonJS modules).

> As of uuid@7 this library now provides ECMAScript modules builds, which allow packagers like Webpack and Rollup to do "tree-shaking" to remove dead code.

PR #231, opened 2020-04-06
Improve geometrie fields

dtrucs:


src/components/react-json-schemaForm/GeometryField/GeometryField.js
```diff
@@ -46,11 +46,12 @@ const resetStyle = (map, layerPaints, name) => {
  */
 const getGeometries = ({
   feature,
-  formData: { coordinates, type },
+  formData,
   params: { id, layer },
   name,
   settings,
 }) => {
+  const { coordinates = [], type } = formData || {};
```
mabhub, line 54: Why extracting `formData` instead of destructuring it in function arguments? Do we forward `formData` anywhere?

dtrucs: No we don't.
But in some cases `formData` could be equals to an object or `null`.  
In the second case, I don't know how we can destructure without extracting.

mabhub: Ok so as I understand, you want to manage `null` like  ̀undefined`. 

This way, you're right, argument destructuring allows fallback value only if argument is `undefined`, when  `||` operator also work for any _falsy_ value, including `null`.

PR #237, opened 2020-04-14
Fix native geolayer selector value conversion

jrmi:


src/modules/RA/Scene/components/GeolayerSelect.js
```diff
@@ -69,8 +69,8 @@ const GeolayerSelect = ({ dataProvider, onChange, excludeIds = [], includeIds =
 
   const handleChoice = ({ target: { value } }) =>
     onChange({
-      geolayer: value,
-      label: geolayers.find(({ id }) => (value === id)).name,
+      geolayer: +value,
+      label: geolayers.find(({ id }) => (+value === id)).name,
     });
```
dtrucs, line 74: If i'm right, this function is a select handle change and to retrieve the label you're building the same loop as you built the select tag. If this is the cas, you should have this information in the event target.

```
  const handleChoice = ({ target, target: { value, options } }) =>
    onChange({
      geolayer: +value,
      label: target[options.selectedIndex].textContent,
    });
```

src/modules/RA/Scene/components/GeolayerSelect.js
```diff
@@ -69,8 +69,8 @@ const GeolayerSelect = ({ dataProvider, onChange, excludeIds = [], includeIds =
 
   const handleChoice = ({ target: { value } }) =>
     onChange({
-      geolayer: value,
-      label: geolayers.find(({ id }) => (value === id)).name,
+      geolayer: +value,
```
dtrucs, line 72: I personally prefer the explicit `Number(value)` but it's the same

src/modules/RA/Scene/components/GeolayerSelect.js
jrmi: Not working with non native select.

src/modules/RA/Scene/components/GeolayerSelect.js
jrmi: Done.

PR #238, opened 2020-04-20
Display attachments for CRUD

dtrucs:


src/modules/CRUD/components/Details/Read/AttachmentView/Attachment/Attachment.js
```diff
@@ -0,0 +1,27 @@
...
+const Attachment = ({ className, file, legend }) => {
+  if (!file) {
+    return null;
+  }
+  return (
+    <div className={className}>
+      <a rel="noopener noreferrer download" target="_blank" href={file}>{legend}</a>
+    </div>
+  );
+};
```
mabhub, line 13:
```suggestion
const Attachment = ({ file, legend, ...props }) => {
  if (!file) {
    return null;
  }
  return (
    <div {...props}>
      <a rel="noopener noreferrer download" target="_blank" href={file}>{legend}</a>
    </div>
  );
};
```

src/modules/CRUD/components/Details/Read/AttachmentView/AttachmentList/AttachmentList.js
```diff
@@ -0,0 +1,40 @@
...
+
+import './styles.scss';
+
+const AttachmentList = ({
+  name,
+  attachments,
+}) => {
+  const Component = name === 'attachments' ? Attachment : Picture;
+  return (
+    <ul className="attachment-list">
```
mabhub, line 15:
```suggestion
const AttachmentList = ({
  name,
  attachments,
  className = '',
  ...props
}) => {
  const Component = name === 'attachments' ? Attachment : Picture;
  return (
    <ul className={`attachment-list ${className}`} {...props}>
```

ou encore :

```suggestion
const AttachmentList = ({
  name,
  attachments,
  className = '',
  ...props
}) => {
  const Component = name === 'attachments' ? Attachment : Picture;
  return (
    <ul className={classnames('attachment-list', className)} {...props}>
```

src/modules/CRUD/components/Details/Read/AttachmentView/Picture/Picture.js
```diff
@@ -0,0 +1,27 @@
...
+const Picture = ({ className, legend, thumbnail }) => {
+  if (!thumbnail) {
+    return null;
+  }
+  return (
+    <div className={className}>
+      <img src={thumbnail} alt={legend} />
+    </div>
+  );
+};
```
mabhub, line 13:
```suggestion
const Picture = ({ legend, thumbnail, ...props }) => {
  if (!thumbnail) {
    return null;
  }

  return (
    <div {...props}>
      <img src={thumbnail} alt={legend} />
    </div>
  );
};
```

src/modules/CRUD/components/Details/Read/AttachmentView/AttachmentList/AttachmentList.js
dtrucs: I don't undestand this suggestion, this ul tag doesn't need a dynamic className

src/modules/CRUD/components/Details/Read/AttachmentView/Attachment/Attachment.js
dtrucs: At the beginning, i also forwarded props on div, but 
1/ I don't need other props on this div, but `className`
2/ It's set a lot of unecessary props on this node like the following example  
`<div class="attachment-item__file" id="4" category="1" action_url="/api/crud/features/b40312d8-6391-4659-9d84-e64a088c2ba2/attachments/4/" created_at="16-04-2020 12:21" updated_at="17-04-2020 17:33">`  
(vs `<div class="attachment-item__file">`)

src/modules/CRUD/components/Details/Read/AttachmentView/Picture/Picture.js
dtrucs: same as https://github.com/Terralego/terra-admin/pull/238#discussion_r411335712

src/modules/CRUD/components/Details/Read/AttachmentView/AttachmentList/AttachmentList.js
mabhub: It is not about a need, but about a global philosophy (best practice) of creating component:
Whenever you use any (potentially external) component that will generate a DOM node, you may want to add any html attribute. At least for debugging purpose.

Forwarding unused non-specific props on sub-element is a way to preserve these expectations, and keep a clean component extendability & reusability without breaking it's own behaviors.

This allows kind of:
```jsx
<AttachmentList style={{ border: '1px solid red' }} />
<AttachmentList style={{ listStyleType: 'none' }} />
<AttachmentList onClick={() => console.log('I\'m here')} />
<AttachmentList className={isInViewport ? 'in-viewport' : 'out-of-sight'} />
<AttachmentList id="myCustomId" />
<AttachmentList id="htmlAnchor1" />
```


@jrmi any thought on this ?

src/modules/CRUD/components/Details/Read/AttachmentView/Attachment/Attachment.js
mabhub: I think one shouldn't pass to a component, props that ain't used by the component nor any child component.

Discussion subject seems to be the same as https://github.com/Terralego/terra-admin/pull/238#discussion_r411347577

dtrucs: True, but these props come from an API response i store. I am not sure it's efficient to filter these each time i used it. 

Moreover, in the future i should use some of these (action_url, id, and category) to CRUDify each item.

src/modules/CRUD/components/Details/Read/AttachmentView/AttachmentView.js
```diff
@@ -1,131 +1,37 @@
...
+import CategoryList from './CategoryList';
+import './styles.scss';
 
-  const list = buildAttachmentList(categories, attachment[section]);
 
-  if (!list) {
-    return <Loading spinner />;
-  }
+const AttachmentView = props => {
+  const {
```
jrmi, line 12: Why don't you destructure the props param instead ?

src/modules/CRUD/components/Details/Read/AttachmentView/Picture/Picture.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Picture = ({ className, legend, thumbnail }) => {
```
jrmi, line 4: Naming : `legend` -> `label` ?

src/modules/CRUD/components/Details/Read/AttachmentView/Attachment/Attachment.js
jrmi: I agree with the idea to try to filter what we give to a component, I'm not sure this is a big overhead to filters the props each time we get data from API (it's just a de-structuration and we do it all the time) but it definitely make the component more readable and the component use also. You immediately know that the Attachment component use only file and legend props.

It's hard for a reviewer to knows what props is really used when you give a component a bunch of data coming from API. I read too many code like that in others Terralego projects and it's not fun, believe me ;-)

However, this component is interchangeable with another one (L.13 of AttachmenLlist.js), so it may seem more delicate to filter all the properties you don't want if the other component uses them. The example here is "thumbnail" which is only useful for the "Picture" component.

PR #239, opened 2020-04-22
Fix TableField if the uiSchema.items is not set

dtrucs:


src/components/react-json-schemaForm/TableField/Tbody/Tbody.js
```diff
@@ -71,7 +71,7 @@ const Tbody = ({
         <tr key={uuid()} className="tableField__tr">
           {Object.keys(orderedValues).map(key => {
             const { [key]: error = {} } = errorSchema[index] || {};
-            const { 'ui:help': uiHelp, ...uiSchemaItem } = uiSchemaItems[key];
+            const { [key]: { 'ui:help': uiHelp, ...uiSchemaItem } = {} } = uiSchemaItems;
```
mabhub, line 74: You *protect* against `uiSchemaItems === undefined`, but [L38](https://github.com/Terralego/terra-admin/blob/3414a0dae1dc6a548a708bcff856ca9cc818e4f1/src/components/react-json-schemaForm/TableField/Tbody/Tbody.js#L38) is not (protected).

Wouldn't it be better defining default value (`{}`) on `uiSchemaItems` for whole component on [L21](https://github.com/Terralego/terra-admin/blob/3414a0dae1dc6a548a708bcff856ca9cc818e4f1/src/components/react-json-schemaForm/TableField/Tbody/Tbody.js#L21)?

PR #242, opened 2020-04-23
Can edit or delete attachments

dtrucs:


src/modules/CRUD/components/Details/Read/AttachmentView/AttachmentList/AttachmentList.js
```diff
@@ -1,36 +1,80 @@
...
+              updateData={updateData}
             />
-            <figcaption className="attachment-item__figcaption">
-              <span className="attachment-item__figcaption-text">{label}</span>
-            </figcaption>
-          </figure>
-        </li>
-      ))}
+          </li>
+        );
```
mabhub, line 76: Couldn't this be in another (sub)component ? (like `AttachmentItem`)

src/modules/CRUD/components/Details/Read/AttachmentView/Figcaption/Figcaption.js
```diff
@@ -0,0 +1,107 @@
...
+import { saveAttachment } from '../../../../../services/attachments';
+
+const Figcaption = ({
+  category,
+  endpoint,
+  label,
+  updateData,
+  t,
+}) => {
+  const [editMode, setEditMode] = useState(false);
```
mabhub, line 13: Maybe
```jsx
  const [isEditing, setIsEditing] = useState(false);
```

src/modules/CRUD/services/attachments.js
```diff
@@ -0,0 +1,16 @@
+import Api from '@terralego/core/modules/Api';
+import { sanitizeCustomEndpoint } from './utils';
+
+const attachmentEndpoint =  (endpoint, body, method = 'PUT') =>
```
mabhub, line 4:
```suggestion
const attachmentEndpoint = (endpoint, body, method = 'PUT') =>
```

mabhub: Also, I don't think `attachmentEndpoint` is the best name for a function that will in fact do a request.

src/modules/CRUD/services/attachments.js
```diff
@@ -0,0 +1,16 @@
...
+  Api.request(`${sanitizeCustomEndpoint(endpoint)}/`, { method, body });
+
+export const saveAttachment = (endpoint, body, method = 'PUT') =>
+  attachmentEndpoint(endpoint, body, method);
```
mabhub, line 8: Why returning the same function with a default argument value,.. which is also already defined in `attachmentEndpoint`…
If ts is just for aliasing, this is enough:
```suggestion
export const saveAttachment = attachmentEndpoint;
```

src/modules/CRUD/components/Details/Read/AttachmentView/AttachmentList/AttachmentList.js
dtrucs: I just made the split

src/modules/CRUD/components/Details/Read/AttachmentView/DeleteAttachment/DeleteAttachment.js
```diff
@@ -0,0 +1,88 @@
...
+  match: {
+    params: {
+      id: undefined,
+    },
+  },
+  t: () => {},
+  updateData: () => {},
+};
+
+export default DeleteAttachment;
```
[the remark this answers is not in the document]
dtrucs: I'll do the refacto right after this PR. The fact is there is another component similar to this one, and if i make the change now, i think the PR would be too large

src/modules/CRUD/services/attachments.js
```diff
@@ -0,0 +1,15 @@
...
+
+export const saveAttachment = attachmentRequest;
+
+export const deleteAttachment = endpoint =>
+  attachmentRequest(endpoint, {}, 'DELETE');
+
+export default {
+  saveAttachment,
+  deleteAttachment,
+};
```
mabhub, line 15: Why using named exports AND default export? Quite overkill…

*(all named exports may always be imported in a single object with `import * as attachmentMethods from './services/attachments.js';`)*

```suggestion
export const saveAttachment = attachmentRequest;
export const deleteAttachment = endpoint => attachmentRequest(endpoint, {}, 'DELETE');
```

src/modules/CRUD/components/Details/Read/AttachmentView/AttachmentItem/AttachmentItem.js
```diff
@@ -0,0 +1,92 @@
...
+}) => {
+  const Component = name === 'attachments' ? Attachment : Picture;
+
+  return (
+    <Tag
+      className={classnames({
+        'attachment-item': true,
+        'attachment-item--loading': isLoading,
+      }, className)}
+      key={id}
```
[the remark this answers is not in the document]
dtrucs: Yes my bad, a faster copy/paste it think 👍

src/modules/CRUD/components/Details/Read/AttachmentView/AttachmentList/AttachmentList.js
```diff
@@ -1,36 +1,58 @@
...
+            key={id}
+            file={file}
+            label={label}
+            name={name}
+            setDeleting={setDeleting}
+            thumbnail={thumbnail}
+            updateData={updateData}
+          />
+        );
+      })}
```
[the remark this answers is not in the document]
dtrucs: I understand the point, but i am not sure to do two `map` loops for one purpose

src/modules/CRUD/components/Details/Read/AttachmentView/AttachmentList/AttachmentList.js
```diff
@@ -1,36 +1,58 @@
...
-  name,
   attachments,
   className,
+  featureEndpoint,
+  fetchFeature,
+  history,
+  location,
+  match: { params: { id: paramId } },
+  name,
+  staticContext,
```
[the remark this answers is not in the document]
dtrucs: Great idea. 👍   
I think it would be more flexible to create a `withRouterParams` HOC than `withParamsId`.

src/modules/CRUD/services/attachments.js
dtrucs: I don't why, i thought the linter does not accept a file without `export default`.  
It's now fixed. 👍

src/modules/CRUD/components/Details/Read/AttachmentView/AttachmentList/AttachmentList.js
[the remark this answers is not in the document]
dtrucs: I finnaly use the hook `useParam` from `react-router-dom'. It works like a charm 👍

src/modules/CRUD/components/Details/Read/AttachmentView/AttachmentList/AttachmentList.js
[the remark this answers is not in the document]
mabhub: The purpose is separation of concerns:
- First manipulating data to prepare it (testable/cacheable/pure function), 
- Then displaying the clean data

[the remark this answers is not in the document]
dtrucs: What do you think about exclure props we don't want to spread all others. Something like=

```jsx
  const sanitizeAttachments = useMemo(() => (
    attachments.map(({
      action_url: endpoint,
      legend: label,
      created_at: createdAt,
      updated_at: updatedAt,
      ...rest
    }) => ({
      endpoint,
      label,
      isLoading: rest.id === isDeleting,
      ...rest,
    }
    ))
  ), [attachments, isDeleting]);

  return (
    <ul className={classnames('attachment-list', className)} {...props}>
      {sanitizeAttachments.map(attachmentProps => (
        <AttachmentItem
          className="attachment-list__item"
          key={attachmentProps.id}
          name={name}
          setDeleting={setDeleting}
          updateData={updateData}
          {...attachmentsProps}
        />
      ))}
    </ul>
  );
```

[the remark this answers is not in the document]
dtrucs: I just pushed this version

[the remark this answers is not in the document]
mabhub: If you think that attachment object must be extendable and that each new object property should be forward to `AttachmentItem` component as prop, that is indeed the way.

If no other prop that those already defined must be passed to component, maybe cherry-picking only the good ones (as I suggested) is better.

*(I've no preference. It is just a matter of object semantic)*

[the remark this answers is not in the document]
dtrucs: I think it's fine in that way 👍

src/modules/CRUD/components/Details/Read/AttachmentView/AttachmentList/AttachmentList.js
```diff
@@ -1,35 +1,55 @@
...
+      created_at: createdAt,
+      updated_at: updatedAt,
+      ...rest
+    }) => ({
+      endpoint,
+      label,
+      isLoading: rest.id === isDeleting,
+      ...rest,
+    }))
+  ), [attachments, isDeleting]);
```
mabhub, line 40: As `sanitizeAttachments` is in component scope, other (common) props may also be forward as object properties:

```suggestion
  const commonProps = { name, setDeleting, updateData };

  const sanitizeAttachments = useMemo(() => (
    attachments.map(({
      action_url: endpoint,
      legend: label,
      created_at: createdAt,
      updated_at: updatedAt,
      ...rest
    }) => ({
      endpoint,
      label,
      isLoading: rest.id === isDeleting,
      key: rest.id,
      ...commonProps,
      ...rest,
    }))
  ), [attachments, isDeleting, commonProps]);
```

then:

```jsx
        <AttachmentItem className="attachment-list__item" {...attachmentProps} />
```

dtrucs: Ok fine :D

dtrucs: Ok fine :D

PR #245, opened 2020-04-24
Add import file feature

dtrucs:


src/modules/CRUD/components/Details/Read/AttachmentView/ImportFile/ImportFile.js
```diff
@@ -0,0 +1,231 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Button, Intent } from '@blueprintjs/core';
+import Uppy from '@uppy/core';
+import { Dashboard } from '@uppy/react';
+import { buildHeaders } from '@terralego/core/modules/Api';
+import locale from '@uppy/locales/src/fr_FR';
```
dtrucs, line 7: I need some advice for this point because I have an i18n issue.
We can get the language code with react-i18next but I don't really know how to dynamically import the right language.
An idea could be to wrap `this.uppy` in the constructor with a dynamic import like the following example:

```jsx
import(`@uppy/locales/src/${lng}`).then(locale => {
  this.uppy = Uppy({
    locale,
    // etc..
  });
})
```

But if the value of `lng` doest not exist in `@uppy/locales/src/` the app crashes immediately.
It's too dangerous.

Need some help.

mabhub: How many locales should we support ?

If we just need fr/en, as Uppy locales are currently not so big (~5 Ko), web could maybe import both in a single object then provide only the good one to component (falling back to _en_ if unavailable).

```js
import localeFr from '@uppy/locales/src/fr_FR';
import localeEn from '@uppy/locales/src/en_US';

const locales = {
  fr: localeFr,
  en: localeEn,
  default: localeEn,
};
```

then in component:

```jsx
    const locale = locales[locales[lng] ? lng : 'default'];
```


You may also simply use the exemple you suggest but with adding a simple check:
```jsx
const UPPY_LOCALES = ['fr_FR', 'en_US', /* … */];
const validLng = UPPY_LOCALES.includes(lng) ? lng : 'en_EN';

import(`@uppy/locales/src/${validLng}`).then(locale => {
  this.uppy = Uppy({
    locale,
    // etc..
  });
})
```

src/modules/CRUD/components/Details/Read/AttachmentView/ImportFile/ImportFile.js
```diff
@@ -0,0 +1,231 @@
...
+
+import UppySetCategoryBeforeUpload from './UppySetCategoryBeforeUpload';
+import CategorySelector from '../CategorySelector';
+
+import '@uppy/core/dist/style.css';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import '@uppy/dashboard/dist/style.css';
+import './styles.scss';
+
+class ImportFile extends React.Component {
```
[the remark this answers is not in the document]
dtrucs: Because it's the recommended way by the package https://uppy.io/docs/react/dashboard/#Initializing-Uppy

> A simple approach is to initialize it in your React component’s constructor() and destroy it in componentWillUnmount().

Anyway, the last week i read this PR, maybe we could convert it in functional component and use `useMemo` https://github.com/transloadit/uppy/pull/1936

src/modules/CRUD/components/Details/Read/AttachmentView/ImportFile/ImportFile.js
dtrucs: This seems a good approach, I chose the second.

src/modules/CRUD/components/Details/Read/AttachmentView/ImportFile/UppySetCategoryBeforeUpload.js
```diff
@@ -0,0 +1,54 @@
...
+  install () {
+    this.uppy.addPreProcessor(this.setCategoryBeforeUpload);
+  }
+
+  uninstall () {
+    this.uppy.removePreProcessor(this.setCategoryBeforeUpload);
+  }
+}
+
+export default UppySetCategoryBeforeUpload;
```
mabhub, line 54: As this seems to be a plugin for Uppy, and has nothing in common with a component (nor a service dedicated to a component), shouldn't this be put appart in a dedicated `lib`-like directory, outside of `/components` directory?

mabhub: Also, couldn't this be contributed as a dedicated plain Uppy plugin?

dtrucs: In fact, it's very specific for the usecase. For a better understanding, below my approach ⬇️ 

My goal was to set category to files before uploading them.

The endpoint to POST a new attachment has a required category ID. To be able to submit files, the user must choose a category. Once selected, and then clicked on the upload button, there are two cases:
* The category already exists so we set its related ID.
* The category does not exist and we POST the new category and await the response of the call to set the new ID.

 `uppy` has a method `onBeforeUpload` to do this kind of manipulation, but it does not wait for asynchronous function https://uppy.io/docs/uppy/#onBeforeUpload-files-gt-files

>Note that this method is intended for quick synchronous checks/modifications only. If you need to do an async API call, or heavy work on a file (like compression or encryption), you should utilize a custom plugin instead.

That's why i built this plugin.

mabhub: Good explanation, it still seems like as a good plugin to spread to the community!
This plugin add a feature that might help a lot of other developers.

Maybe not exactly as is: You could take the `setCategoryBeforeUpload` function out of the plugin and provide it as a plugin argument.
Then, the main intent of the plugin would be to allow awaiting an `onBeforeUpload` call.

Doing it this way is not mandatory, but it is clearly a proper business & logic opportunity.

src/modules/CRUD/components/Details/Read/AttachmentView/ImportFile/ImportFile.js
```diff
@@ -0,0 +1,268 @@
...
+import '@uppy/core/dist/style.css';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import '@uppy/dashboard/dist/style.css';
+import './styles.scss';
+
+const locales = {
+  fr: localeFr,
+  en: localeEn,
+  default: localeEn,
+};
```
mabhub, line 25: You may `export` this in another file: 

```js
// uppy-locales.js
import localeFr from '@uppy/locales/src/fr_FR';
import localeEn from '@uppy/locales/src/en_US';

const locales = {
  fr: localeFr,
  en: localeEn,
  default: localeEn,
};

export default locales;
```
then:
```js
// ImportFile.js
import locales from './uppy-locales.js';
```

src/modules/CRUD/components/Details/Read/AttachmentView/ImportFile/index.js
```diff
@@ -0,0 +1,41 @@
...
+}) => {
+  const { featureEndpoint } = getView(settings, layer);
+  return {
+    findOrCreateAttachmentCategory,
+    feature: feature[id],
+    featureEndpoint,
+    fetchFeature,
+    imageAllowedFileTypes,
+    fileAllowedFileTypes: null,
+    maxFileSize: typeof maxFileSize === 'number' ? maxFileSize * 8 : null, // `* 8` is the conversion from octet to bytes
```
mabhub, line 33: You do **not** convert octets to bytes, as octets actually *are* bytes (of 8 bits).

```suggestion
    maxFileSize: typeof maxFileSize === 'number' ? maxFileSize * 8 : null, // `* 8` is the conversion from bytes to bits
```

and maybe more precisely:
```suggestion
    maxFileSize: typeof maxFileSize === 'number' ? maxFileSize * 8 : null, // `* 8` is the conversion from octets to bits
```

src/modules/CRUD/components/Details/Read/AttachmentView/ImportFile/uppy-locales.js
```diff
@@ -0,0 +1,11 @@
+import en from '@uppy/locales/src/en_US';
+import fr from '@uppy/locales/src/fr_FR';
+// More locales here https://github.com/transloadit/uppy/tree/master/packages/%40uppy/locales/src
```
mabhub, line 3:
```suggestion
// See https://uppy.io/docs/locales/#List-of-locale-packs for more locales
```

src/modules/CRUD/components/Details/Read/AttachmentView/ImportFile/ImportFile.js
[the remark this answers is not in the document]
jrmi: It's not the recommended way but a "simple approch". I think the doc was written before hooks and functional component ;-) 
I'll not insist, you can resolve this when you've read it.

PR #247, opened 2020-05-06
Able to pick image attachment from RTE field

dtrucs:


src/components/react-json-schemaForm/RTEField/PictureSelector/PictureSelector.js
```diff
@@ -0,0 +1,56 @@
...
+    <Overlay
+      className={Classes.OVERLAY_SCROLL_CONTAINER}
+      elevation={Elevation.TWO}
+      isOpen
+    >
+      <div className="picture-selector">
+        <Card>
+          <header className="picture-selector__header">
+            <H5>{t('jsonSchema.RTEField.pick-image')}</H5>
+            <Button minimal icon="small-cross" title="Fermer" onClick={onClose} />
```
jrmi, line 23: The title is in french.

dtrucs: Good catch 👍 
Fixed.

src/modules/CRUD/components/Details/Read/AttachmentView/CategoryList/CategoryList.js
```diff
@@ -5,7 +5,7 @@ import Message from '../../../../../../../components/Message';
 
 import AttachmentList from '../AttachmentList';
 
-const CategoryList = ({ categories, name, t }) => {
+const CategoryList = ({ categories, editable, name, onSelection, selectable, t }) => {
```
mabhub, line 8:
```suggestion
const CategoryList = ({ categories,  name, t, ...rest }) => {
```

then:

```jsx
          <AttachmentList
            attachments={attachments}
            name={name}
            {...rest}
          /> 
```

dtrucs: I like it, but this component includes `withRouter` and `withTranslation` HOCs.   
With your suggestion, it should forward a lot of useless props and problematical props such as `staticContext`.

mabhub: Is see…
But there's no real reason for `CategoryList` to be wrapped by `withRouter` _(as it does not use any routing props)_, but I just saw that's for being able having `connectCRUDProvider` to use routing information.

A solution could be having an HOC dedicated to strip props that are useless for the component. I not sure it is the best solution, but I think it is still be better than forwarding useless props:

```jsx
const stripProps = WrappedComponent =>
    ({
        // props to remove
        staticContext,
        ...props
    }) => <WrappedComponent {...props} />;

export default compose(
  withRouter,
  withTranslation(),
  connectCRUDProvider(CRUDPRoviderGetter),
  stripProps,
)(CategoryList);
```

*(@jrmi any thought on this?)*

mabhub: *(We should always think that components we create should know if it's wrapped or not by an HOC. Also, it is important to keep extendability, and forwarding props should **not** be an issue.)*

jrmi: I would have agreed if it was a basic tag but for a component I prefer the explicit as much as possible, it seems more readable to me. Which gives :

```suggestion
const CategoryList = ({ categories, name, selectable, ...rest }) => {
```

Then

```jsx
const {editable, onSelection, selectable} = rest;

// […]

<AttachmentList
        attachments={attachments}
        name={name}
        editable={editable}
        onSelection={onSelection}
        selectable={selectable}
    /> 
```

However, I agree with always considering that components may not be decorated by HOCs.

jrmi: Also, I agree with the props cleaning in the HOC stack composition to avoid problems.

dtrucs: I just pushed a commit which sanitize props directly in CategoryList HOC

PR #248, opened 2020-05-06
Refacto confirm deletion

dtrucs:
@mabhub Finally i did this! :D


dtrucs:
> I think that it could easily (without much effort) be a generic `ActionConfirm` (instead of `ConfirmDeletion`), but that is already clearly better than previous _to much specialized_ `DeleteFeature`.

Right i though about it. But i think a refacto could be done if we need a different type of confirm than deletion

PR #250, opened 2020-05-07
Add fitbound button in details header

dtrucs:


src/modules/CRUD/components/Details/Header/Header.js
```diff
@@ -1,16 +1,20 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import FitBoundButton from '../FitBoundButton';
 import DownloadButtons from '../DownloadButtons';
 
 const Header = ({ title, documents }) => (
-  <div className="details__header">
-    <h2 className="details__title">{title}</h2>
+  <header className="details__header">
```
mabhub, line 7: I would have taken the opportunity to make this component extensible by allowing arbitrary props to be applied on `<header />`.

*(Not mandatory at all!)*

src/modules/CRUD/components/Details/FitBoundButton/FitBoundButton.js
```diff
@@ -0,0 +1,54 @@
+import React, { useCallback } from 'react';
+import PropTypes from 'prop-types';
+import { Button } from '@blueprintjs/core';
+
+import { getBounds } from '../../../services/features';
+
+const FitBoundButton = ({ coordinates, detailsRef, map, t, title }) => {
+  const handleClick = useCallback(() => {
+    const { current: detail } = detailsRef;
+    if (!map && !detail) {
```
mabhub, line 10: I think this should be a OR:

```suggestion
    if (!map || !detail) {
```

src/modules/CRUD/components/Details/Header/Header.js
dtrucs: Ok it's a quick win, i pushed a newer version of it

PR #281, opened 2020-09-21
Upgrade react admin to v3 (again)

mabhub:


src/modules/RA/DataLayer/components/FieldUpdater.js
```diff
@@ -1,13 +1,10 @@
 import { useEffect } from 'react';
-import get from 'lodash.get';
 /* eslint-disable import/no-extraneous-dependencies */
```
dtrucs, line 2: `es-lint disable` is still accurate?

src/modules/RA/DataSource/components/DataSourceCSVFields.js
```diff
@@ -179,7 +179,10 @@ const DataSourceCSVFields = ({ translate: t, type, ...props }) => (
       step={1}
     />
 
-    <h4>{t('datasource.form.optionsCSV.options')}</h4>
+    <>
+      <h4>{t('datasource.form.optionsCSV.options')}</h4>
+    </>
```
dtrucs, line 184: I don't see the utility of this diff, why did you add a fragment tag?

src/modules/RA/Scene/components/SceneFormNameField.js
```diff
@@ -0,0 +1,33 @@
...
+  const form = useForm();
+
+  return (
+    <FormDataConsumer formClassName={formClassName}>
+      {({ formData, ...rest }) => (
+        <TextInput
+          source="name"
+          label="view.form.name"
+          onChange={value => {
+            if (!value) return;
```
dtrucs, line 22: The early return seems unnecessary because `value` is the `onChange` event and i think it should not be falsy

src/modules/RA/DataLayer/components/DataLayerTabbedForm.js
```diff
@@ -53,276 +52,300 @@ const LazyFormTab = ({ hidden, ...props }) => (
...
+            label="datalayer.form.name"
+            validate={defaultRequired}
+            type="text"
+          />
+
+          <FormDataConsumer>
+            {({ formData }) => {
+              const hasFields = formData.fields && formData.fields.length;
+              if (!hasFields) {
+                return <></>;
```
dtrucs, line 133:
```suggestion
                return null;
```

src/modules/RA/DataLayer/components/DataLayerTabbedForm.js
```diff
@@ -19,8 +17,9 @@ import { ColorInput } from 'react-admin-color-input';
 
 /* eslint-disable import/no-extraneous-dependencies */
 import { withStyles } from '@material-ui/core/styles';
-import { change } from 'redux-form';
+// import { change } from 'redux-form';
```
dtrucs, line 20: Line to remove?

src/modules/RA/DataLayer/components/DataLayerTabbedForm.js
```diff
@@ -53,276 +52,300 @@ const LazyFormTab = ({ hidden, ...props }) => (
...
+                                { id: 'number', name: 'datalayer.form.type.range_format.number' },
+                                { id: 'date', name: 'datalayer.form.type.range_format.date' },
+                              ]}
+                            />
+                          )}
+                        </>
+                      );
+                    }}
+                  </FormDataConsumer>
+                  {formData.table_enable ? <BooleanInput source="shown" label="datalayer.form.show" /> : <></>}
```
dtrucs, line 327:
```suggestion
                  {formData.table_enable && <BooleanInput source="shown" label="datalayer.form.show" />}
```

src/modules/RA/DataLayer/components/DataLayerTabbedForm.js
```diff
@@ -53,276 +52,300 @@ const LazyFormTab = ({ hidden, ...props }) => (
...
+                  <FormDataConsumer>
+                    {({ scopedFormData = {}, getSource }) => (
+                      scopedFormData.shown && <BooleanInput source={getSource('display')} label="Afficher ce champs par défaut" />
+                    )}
+                  </FormDataConsumer>
+                  {formData.table_export_enable ? <BooleanInput source="exportable" label="datalayer.form.exportable" /> : <></>}
```
dtrucs, line 333:
```suggestion
                  {formData.table_export_enable && <BooleanInput source="exportable" label="datalayer.form.exportable" />}
```

dtrucs (changes requested):

src/modules/RA/DataSource/components/DataSourceCSVFields.js
jrmi: To protect this element from react-admin props I assume ?

src/modules/RA/DataLayer/components/DataLayerTabbedForm.js
```diff
@@ -53,276 +52,300 @@ const LazyFormTab = ({ hidden, ...props }) => (
...
+        form.change('table_export_enable', false);
+      }}
+    />
+  );
+};
+
+const SourceField = ({ formData, dataProvider }) => {
+  const form = useForm();
+  return (
+    <>
```
jrmi, line 73: Do we need this wrapper ?

src/modules/RA/DataLayer/components/FieldUpdater.js
```diff
@@ -61,22 +58,26 @@ export const updateFieldFromSource = async (layerFields, dispatch, dataProvider,
...
+  const { input: { value: layerFields } } = useField('fields');
+
   // Update fields from source
   useEffect(() => {
     if (!sourceId || !layerFields) return;
 
     const fillFields = async () => {
-      // We should avoid dipatch call if component is unmount
-      await updateFieldFromSource(layerFields, dispatch, dataProvider, sourceId);
+      // We should avoid dispatch call if component is unmount
```
jrmi, line 75:
```suggestion
      // We should avoid update call if component is unmount
```

src/modules/RA/DataLayer/components/FieldUpdater.js
mabhub: I think so.

`react-final-form` is currently defined as root dependency, but I think it should be removed from package.json and be brought as `react-admin` sub-dependency.

Currently: 
```sh
$ npm ls react-final-form
terra-admin@2.14.1 ~/projets/Terralego/terra-admin
├─┬ react-admin@3.8.5
│ └── react-final-form@6.5.1 
└── react-final-form@6.5.1 
```

mabhub: As discussed with @jrmi, we'll try to define `react-final-form` *(and @material-ui/core)* as peerDepedency, to avoid having de setup so much linting exceptions.

mabhub: `react-final-form` is already a peerDep of react-admin so we require it as root dependency.  
Linting manual disabling is so not needed anymore.

src/modules/RA/DataSource/components/DataSourceCSVFields.js
mabhub: > To protect this element from react-admin props I assume ?

Precisely.

*react-admin component do not "just" render its children components. It iterate through them and provides each some props. Many of these props aren't valid for DOM elements. Wrapping with a Fragment make it the receiver of the props, keeping safe the DOM element.*

src/modules/RA/Scene/components/SceneFormNameField.js
mabhub: In some (edge)cases, `onChange` might formerly be called with no argument, making `value.target.value`  to throw an error… 

I don't really know if it could still happens, but In case…

src/modules/RA/DataLayer/components/DataLayerTabbedForm.js
mabhub: Not really in scope of the `react-admin` v2 ➡ v3 migration.

Anyway, `<DraggableFormIterator />` does not accept `null` as child *(it requires strictly a component as child, and `null`is not)*. It may be patched to avoid error when it happens.

mabhub: Also in some cases, null is not a valid for child for some specific components. Returning an empty Fragment is then a quite simple security measure.

src/modules/RA/DataLayer/components/DataLayerTabbedForm.js
mabhub: Not anymore *(as `FieldUpdater is now inside a hidden tab)*

dtrucs (approved):

jrmi (approved):

PR #283, opened 2020-09-23
Cut off dead code from old OPP admin UI

mabhub:
As far as I know, this `OPP` "module" *(arbitrary name given to a delimited set for features that define a "part" of the main application)* is not use anymore in anyway.

All features and requirements of *Observatoire Photographique des Paysage* should have been already rebuild entirely in `RA` "module".


jrmi (approved):

dtrucs:
Updates to translation files are missing.  
At first glance it looks like everything is in the `opp` key. 
https://github.com/Terralego/terra-admin/blob/master/public/locales/fr/translation.json#L98

PR #293, opened 2020-10-05
Add export geomfiles feature

dtrucs:


src/modules/CRUD/components/Details/ExportGeomFiles/ExportGeomFiles.js
```diff
@@ -0,0 +1,58 @@
...
+  const list = Object.entries(files);
+  const { t } = useTranslation();
+
+  const count = list.length;
+
+  if (!count) {
+    return null;
+  }
+
+  const handleClick = async event => {
```
jrmi, line 19: useCallback maybe ?

src/modules/CRUD/components/Details/ExportGeomFiles/ExportGeomFiles.js
```diff
@@ -0,0 +1,58 @@
...
+    event.preventDefault();
+    const href = event.target.closest('a').getAttribute('href');
```
jrmi, line 21: Not a fan of doing that. Can you set a state with the file and show the element if the state is not null instead ?

jrmi (changes requested):

src/modules/CRUD/components/Details/ExportGeomFiles/ExportGeomFiles.js
mabhub: `Element.closest()` is not supported by IE at all. *(I do not know IE support status of terra-admin)*

image

jrmi: It's not a compat issue for me but more a philosophy question. Should we really do DOM manipulation through js API in a react code ?

dtrucs: Concerning @mabhub issue, i thought `'react-app-polyfill/ie11'` worked with it, but apparently nothing about traversing DOM. So i might replace `event.target.closest('a')` by `event.currentTarget` (which is finally better).

@jrmi would you prefer the pass the href directly as parameter in the `handleClick`?
```jsx
      {list.map(([name, href]) => (
        <MenuItem
          className={Classes.MINIMAL}
          icon="document"
          text={name}
          key={name}
          href={href}
          onClick={handleClick(href)}
        />
      ))}
```
and 
```js

  const handleClick = href => async event => {
    event.preventDefault();
    const data = await Api.request(sanitizeCustomEndpoint(href), { responseType: 'text' });
    const file = new Blob([data], { type: 'text/plain' });

    const [fileName] = href.split('/').slice(-1);
    const link = document.createElement('a');
    link.download = fileName;
    link.href = global.URL.createObjectURL(file);
    link.click();
  };
```

src/modules/CRUD/components/Details/ExportGeomFiles/ExportGeomFiles.js
dtrucs: In this function, I don't manipulate any props or states, do you think it's necessary?

jrmi (approved):

PR #304, opened 2020-10-08
Better filter tab

jrmi:


src/modules/RA/DataLayer/components/DataLayerForm.js
```diff
@@ -94,8 +65,5 @@ const PropsSanitizer = WrappedComponent =>
   ({ withSource, dispatch, ...props }) => (<WrappedComponent {...props} />);
 
 export default compose(
-  withStyles(styles),
-  translateRA,
   PropsSanitizer,
```
mabhub, line 68: Is `PropsSanitizer` still required?

jrmi: No anymore you're right. Removed.

CFarcy (dismissed):

CFarcy (approved):

PR #312, opened 2020-11-03
Add title to popup form

CFarcy:


src/modules/RA/DataLayer/components/tabs/PopupTab/index.js
```diff
@@ -50,24 +60,32 @@ const PopupTab = () => {
...
+      );
+
+      const lines = contentfields.map(({
         suffix,
         prefix,
         field: { name },
         default: defaultTitle,
       }) => (
-        `${prefix}: {% if ${name}.length > 0} %}{{${name}}} ${suffix}{% else %}${defaultTitle}{% endif %}`
+        `- ${prefix} : {% if ${name} %}{{${name}}} ${suffix}{% else %}${defaultTitle}{% endif %}`
```
jrmi, line 76: 0 values are defined and shouldn't show default message.

jrmi: Also the prefix is not the line title. the prefix is just before the value. Use the field name instead.

jrmi: Handle value round.

jrmi (changes requested):

src/modules/RA/DataLayer/components/tabs/PopupTab/index.js
CFarcy: Done.

src/modules/RA/DataLayer/components/tabs/PopupTab/index.js
```diff
@@ -47,27 +57,49 @@ const PopupTab = () => {
...
         default: defaultTitle,
-      }) => (
-        `${prefix}: {% if ${name}.length > 0} %}{{${name}}} ${suffix}{% else %}${defaultTitle}{% endif %}`
-      ));
+        sourceFieldId,
+      }) => {
+        const { settings: { round } } = fields.find(f => (f.sourceFieldId === sourceFieldId));
+        if (round !== undefined) {
+          return (
+            `- ${label} : ${prefix} `
```
jrmi, line 79: Prefix part should be inside if, like the suffix.

jrmi (dismissed):

src/modules/RA/DataLayer/components/tabs/PopupTab/index.js
CFarcy: Done.

jrmi (approved):

PR #327, opened 2020-12-10
Minisheet form stabilization

CFarcy:


public/locales/en/translation.json
```diff
@@ -540,7 +541,8 @@
           "round": "Number of digit to round"
         },
         "true": "Yes",
-        "false": "No"
+        "false": "No",
+        "color-label": "Select a color"
```
jrmi, line 545: A color for what ?

CFarcy: For highlight! I will precise it.

jrmi (approved):

PR #333, opened 2020-12-15
Popup & MiniSheet stabilization

CFarcy:


src/components/react-admin/ColorPicker.js
```diff
@@ -40,7 +40,7 @@ const useStyles = makeStyles({
 
 const popover = {
   position: 'absolute',
-  zIndex: '2',
+  zIndex: '11',
```
jrmi, line 43: This is a generic component, you shouldn't modify zIndex here

CFarcy: Alright. Will modify elsewhere then!

src/modules/RA/DataLayer/components/tabs/PopupTab/PopupFielSelect.js
```diff
@@ -10,16 +10,17 @@ const FieldSelect = ({
   path,
   fields,
   selectable,
-  selected,
+  selected = '',
```
jrmi, line 13: `selected` should be null or number ?

CFarcy: I avoided null. In this component, it will be use as a value for the `select` and I want the component to stay "controlled".

src/components/react-admin/ColorPicker.js
CFarcy: Done.

jrmi (approved):

PR #345, opened 2021-01-20
Add new fields in user form for further infos

CFarcy:


src/modules/RA/User/components/UserFields.js
```diff
@@ -1,35 +1,78 @@
...
+  return (
+    <SimpleForm {...props}>
+      {edit && <TextInput disabled source="id" />}
+      {edit && <TextInput disabled source="uuid" />}
+
+      <TextInput source="properties.first_name" label="user.form.firstname" />
+      <TextInput source="properties.last_name" label="user.form.lastname" />
+      <TextInput source="email" type="email" />
+      {!edit && <TextInput source="password" type="password" />}
+      <BooleanInput source="is_superuser" />
```
dtrucs, line 39: Maybe it should have a condition to display/control this boolean? Such as only superUser can access/edit this setting

CFarcy: Real question: User creation shouldn't it be accessible to superUser only?

jrmi: User creation should be accessible to user with "user_management" permissions only. That the point of permissions field that list functional permissions that imply available modules. 
Forbidden modules shouldn't be displayed to user. For now, if a user can manage users, he can create super users. May be later we'll make something to protect that.

jrmi (approved):

jrmi (dismissed):

src/modules/RA/User/components/UserFields.js
```diff
@@ -1,35 +1,79 @@
...
+  return (
+    <SimpleForm {...props}>
+      {edit && <TextInput disabled source="id" />}
+      {edit && <TextInput disabled source="uuid" />}
+
+      <TextInput source="properties.first_name" label="user.form.firstname" />
+      <TextInput source="properties.last_name" label="user.form.lastname" />
+      <TextInput source="email" type="email" />
+      {!edit && <TextInput source="password" type="password" />}
+      { isSuperUser && <BooleanInput source="is_superuser" />}
```
dtrucs, line 40: It looks good to me!
Maybe it's better to display the field but disabled in order to escort the user. Something like this ⬇️ 
```jsx
<BooleanInput 
  source="is_superuser" 
  disabled={!isSuperUser}
  helperText={!isSuperUser && 'You don't have enough permission to edit this field' }
/>
```
 

```

src/modules/RA/User/components/UserFields.js
jrmi: We discuss this IRL !

dtrucs (approved):

src/modules/RA/User/components/UserFields.js
CFarcy: Done!

dtrucs (approved):

PR #367, opened 2021-02-15
Force correct value on advanced value for minisheet

CFarcy:


src/modules/RA/DataLayer/components/tabs/MinisheetTab/index.js
```diff
@@ -50,7 +50,10 @@ const MinisheetTab = () => {
   const { geom_type: geomType, id: sourceId } = useSourceData('source');
 
   const { input: { value: fields } } = useField('fields');
-  const { input: { value: advanced } } = useField('minisheet_config.advanced', { initialValue: true });
+  const { input: { value: advanced } } = useField('minisheet_config.advanced', {
+    initialValue: true,
+    format: v => Boolean(v),
```
dtrucs, line 55: Just thinking about it but
```jsx
    format: Boolean,
```
should work, doesn't it?

CFarcy: Done.

jrmi (approved):

PR #381, opened 2021-02-24
Disabling viewpoint

CFarcy:


src/modules/RA/Viewpoint/components/ViewpointFields.js
```diff
@@ -149,6 +149,7 @@ const ViewpointFields = ({
       {...props}
     >
       <FormTab label="resources.viewpoint.tabs.general">
+        <BooleanInput source="active" />
```
jrmi, line 152: Is it easy to add something to show that the viewpoint is not visible in the list display ?

jrmi: Like a stroked eye icon on top of the image

CFarcy: Yes, good idea.

CFarcy: Done.

jrmi (approved):

PR #382, opened 2021-02-25
rework minisheet to handle partial data retrieved from RA

CFarcy:


src/components/react-admin/PreventPartialData.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+/**
+ * Data MUST always have the same shape
```
jrmi, line 4:
```suggestion
 * In react-admin, data MUST always have the same shape. see:
```

src/modules/RA/DataLayer/components/tabs/MinisheetTab/MiniSheetFieldTree.js
```diff
@@ -88,6 +89,15 @@ const MiniSheetFieldTree = ({
     };
   }, [fields, sourceFieldId]);
 
+  const onTitleChange = useCallback(cb => ({ target: { value } }) => {
```
jrmi, line 92: The useCallback doesn't works here. Remove it or change the way it's working.

CFarcy: Removed.

jrmi (approved):

PR #401, opened 2021-03-17
Change Weather Input of viewpoint from textfield to select

vllorca:


src/modules/RA/Picture/components/PictureFields.js
```diff
@@ -30,6 +30,13 @@ const styles = {
   },
 };
 
+const weatherCondition = [
```
CFarcy, line 33: Maybe adding plural to variable name, `weatherConditions` to be more explicit. Also, french should not be wrote directly in the code base, translated content is prefered.

Since those data will be hard coded here, maybe it is preferable to send more human readable codename instead of numeric id to backend? Something like `good_weather`, `lightly_covered` etc

src/modules/RA/Picture/components/PictureFields.js
```diff
@@ -30,6 +30,13 @@ const styles = {
...
+  { id: 1, name: 'Beau temps' },
+  { id: 2, name: 'Légérement couvert' },
```
jrmi, line 35: Also don't use Id but slug like "good_weather" "cloudy", ....

CFarcy (dismissed):

public/locales/en/translation.json
```diff
@@ -802,6 +802,14 @@
       "email": "Should be a valid email",
       "oneOf": "Should be: %{options}",
       "regex": "Should check a specific format (regexp): %{pattern}"
+    },
+    "pictures": {
+      "weather": {
+        "good_weather": "Good weather",
+        "lightly_covered": "Lighly covered",
```
jrmi, line 809:
```suggestion
        "lightly_covered": "Lighly cloudy",
```

public/locales/en/translation.json
```diff
@@ -802,6 +802,14 @@
...
+        "partialy_covered": "Partialy covered",
```
jrmi, line 810:
```suggestion
        "partialy_covered": "Partialy cloudy",
```

jrmi (changes requested):
Some translations need to be updated.

jrmi (approved):

CFarcy (approved):

jrmi (approved):

PR #403, opened 2021-03-17
Redirection to list after adding new picture

vllorca:


src/modules/RA/Picture/views/Create.js
```diff
@@ -4,13 +4,16 @@ import { Create } from 'react-admin';
 import PictureFields from '../components/PictureFields';
 import DefaultActions from '../../../../components/react-admin/DefaultActions';
 
+const redirect = (basePath, id, data) => `/viewpoint/${data.viewpoint}`;
+
 export const PictureCreate = props => (
   <Create
     {...props}
-    actions={<DefaultActions />}
+    actions={<DefaultActions redirect={redirect} />}
```
jrmi, line 12: Does the redirection works if we don't came from a viewpoints ?

vllorca: Yes it does. I changed the redirection in the Create component
It's the one called when you add a picture, coming from a viewpoint or directly from the picture list.

jrmi: Actually, we want to come back to the viewpoint if we came from it but come back to picture list if we don't. Is this the implemented behaviour ?

vllorca: As describe in the ticket#791, after adding a picture we want to come back to the list of photo from the corresponding viewpoint, and not the pictures list for all the observatory.

image

jrmi: Yes but only when you came from a viewpoint as the first words explain it.

jrmi (approved):

CFarcy (approved):

PR #414, opened 2021-03-30
Add "search location" in map control

dtrucs:


src/modules/CRUD/services/MapProvider.js
```diff
@@ -29,11 +33,95 @@ export const MapProvider = ({ children }) => {
...
+      top: 20,
+      right: hasDetails ? (detail?.offsetWidth ?? 0 + 50) : 50,
+      bottom: !hasDetails ? (dataTable?.offsetHeight ?? 0 + 20) : 20,
+      left: 20,
+    };
+    map.resize();
+    map.fitBounds(getBounds(coordinates), { padding, duration: 0 });
+  }, [dataTableRef, detailsRef, map]);
+
+  const search = useMemo(() => ({
```
jrmi, line 56: Why a useMemo instead of two useCallback ?

dtrucs: Because it's not a function but a constante

jrmi: It's an object with two function properties that are spread in a use effect if I read well ?

dtrucs: Oh i misread 'two'.
I think i coded in this way because at the beginning it was easier.
But now i see what you mean :D.

Do you think it's better to use two useCallack for each method?

dtrucs: Fixed 👍

jrmi (approved):

PR #423, opened 2021-04-01
Hide baselayer module

vllorca:


src/modules/RA/ra-modules.js
```diff
@@ -141,6 +141,7 @@ export const config = {
     {
       label: 'baseLayer.project',
       requiredModule: 'BaseLayer',
+      // requiredPermissions: 'can_manage_baselayers',
```
CFarcy, line 144: Commented code should not be present, unless for very precise reasons

src/modules/RA/ra-modules.js
```diff
@@ -99,8 +99,8 @@ export const resources = [
   {
     name: RES_BASELAYER,
     moduleName: 'BaseLayer',
-    ...baseLayerViews,
     endpoint: 'baselayer',
+    ...baseLayerViews,
```
CFarcy, line 103: What is the reason behind this change?

src/modules/RA/ra-modules.js
vllorca: I kept this comment, as permissions could be a solution for the ticket and should be discussed with Pauline or Jeremie

vllorca: Do you want me to erase this comment ?

src/modules/RA/ra-modules.js
vllorca: I did this for the object to match the the ones in the Array
It does not change the behavior, do you want me to revert this change ?

src/modules/RA/ra-modules.js
```diff
@@ -150,6 +151,11 @@ export const config = {
   ],
 };
 
+if (process.env.REACT_APP_NOT_DISPLAY_BASELAYERS) {
+  const { menu } = config;
+  config.menu = menu.filter(obj => obj.label !== 'baseLayer.project');
+}
+
```
dtrucs, line 158: Maybe you could set a more generic stuff by setting a "pseudo array".
For example in the .env file you've got the following line ⬇️ 
`REACT_APP_NOT_DISPLAY_MODULES="baseLayer foo bar"`

And then in the code you can iterate on it
```js
if (process.env.REACT_APP_NOT_DISPLAY_MODULES) {
  const { menu } = config;
  process.env.REACT_APP_NOT_DISPLAY_MODULES.split(' ').forEach(module => {
   config.menu = menu.filter(({ label }) => label !== module);
  }
}
```

dtrucs: ok i miss `.project` but it's possible with a .map after the split.
What do you think?

src/modules/RA/ra-modules.js
CFarcy: Not  necessarily revert the change, but as it is a stylistic change, would be better in a separate commit in my opinion.

src/modules/RA/ra-modules.js
CFarcy: Even though some change could appear in the work of this PR, commented code should not be present.

src/modules/RA/ra-modules.js
CFarcy: I like this idea

vllorca: I applied the changes
There is now a env variable looking like REACT_APP_NOT_DISPLAY_MODULES="baseLayer.project foo.project bar.project"

dtrucs: Yes top!
Maybe you can avoid the repetition of '.project' in the env file and match the correspondance here
```js
const notDisplayModulesProject = !process.env.REACT_APP_NOT_DISPLAY_MODULES.split(' ').map(module => `${module}.project`}
config.menu = menu.filter(({ label }) => notDisplayModulesProject.includes(label));`
```

or but maybe less accurate
```js
config.menu = menu.filter(({ label }) => !process.env.REACT_APP_NOT_DISPLAY_MODULES.split(' ').includes(label.replace('.project', ''));
```

vllorca:
This issue was solved in django-terra-account using a django manager to filter modules and permissions with a disabled_module setting

PR #425, opened 2021-04-07
Display identifier instead of id when deleting picture

vllorca:


src/components/react-admin/DeleteWithConfirmButtonJS.js
```diff
@@ -0,0 +1,101 @@
...
+    icon = defaultIcon,
+    label = 'ra.action.delete',
+    mutationMode,
+    onClick,
+    record,
+    redirect = 'list',
+    ressource,
+    onSuccess,
+    onFailure,
+    idFields,
```
jrmi, line 47: What if we want to use 2 fields in our message? Not important for our needs, but for a PR on React-admin, should be more reusable to have something like `translateOptions` direct overiding? What do you think??

vllorca: I'll try passing `translateOptions` as props instead of IdFields

src/components/react-admin/DeleteWithConfirmButtonJS.js
```diff
@@ -0,0 +1,97 @@
...
+    confirmContent = 'ra.message.delete_content',
+    icon = defaultIcon,
+    label = 'ra.action.delete',
+    mutationMode,
+    onClick,
+    record,
+    redirect = 'list',
+    onSuccess,
+    onFailure,
+    translateOptions,
```
jrmi, line 46: Can you add the previous default ?

jrmi (dismissed):

src/components/react-admin/DeleteWithConfirmButtonJS.js
```diff
@@ -0,0 +1,115 @@
...
+import {
+  Button,
+  Confirm,
+  useDeleteWithConfirmController,
+  useResourceContext,
+  useTranslate,
+} from 'react-admin';
+
+import classnames from 'classnames';
+import inflection from 'inflection';
```
dtrucs, line 15: Where is `inflection` package? 
Did you forget to commit its installation in the package.json (and lock)?

jrmi (dismissed):

src/components/react-admin/DeleteWithConfirmButtonJS.js
vllorca: I did it

jrmi (approved):

CFarcy (approved):

PR #429, opened 2021-04-12
Fix name in delete with confirm button

vllorca:


public/locales/en/translation.json
```diff
@@ -968,7 +968,8 @@
       "tabs": {
         "metadata": "Meta data",
         "repeat": "Rephotography"
-      }
+      },
+      "name": "Photographie"
```
CFarcy, line 972: Becarefull, this is the english `translation` file

vllorca: Thanks, must be a copy paste mistake

jrmi (approved):

PR #432, opened 2021-04-13
Refacto geometry field and add reset feature

dtrucs:


src/components/react-json-schemaForm/GeometryField/Informations.js
```diff
@@ -13,26 +15,34 @@ const getIconFromType = type => {
...
+    nextFormData: { geom } = {},
+  } = useContext(GeometryFieldContext);
+
 
   const { t } = useTranslation();
 
   const { coordinates = schemaCoordinates, type = schemaType } = geom || {};
 
-  const action = !coordinates.length ? 'create' : 'edit';
+  const hasDraws = Boolean(coordinates.length);
```
jrmi, line 35:
```suggestion
  const hasDraw = Boolean(coordinates.length);
```

src/components/react-json-schemaForm/GeometryField/Informations.js
```diff
@@ -13,26 +15,34 @@ const getIconFromType = type => {
...
-    formData: { geom } = {},
     isRequired,
     isRouting,
-    schema: { default: {
-      coordinates: schemaCoordinates,
-      type: schemaType,
-    } },
-  } = props;
+    nextFormData: { geom } = {},
+  } = useContext(GeometryFieldContext);
```
jrmi, line 28: You can add a `useGeometryField` hook and use it instead. It's a little more clean (only a minor detail)

dtrucs: Yes, totally right!
Done 👍

jrmi (approved):

PR #435, opened 2021-04-14
Standardize case for cities

vllorca:


CFarcy (approved):

src/components/react-admin/FreeAutocompleteInput.js
```diff
@@ -17,13 +17,16 @@ const FreeAutocompleteInput = ({ choices = [], label, ...rest }) => {
     meta: { touched, error },
   } = useInput(rest);
 
+  const lowercasedChoices = choices.map(choice =>
```
jrmi, line 20: Should be done backend side ;-)

vllorca:
The issue was solved in terra-opp

PR #443, opened 2021-04-21
Add confirmations to picture actions

jrmi:


vllorca (approved):

CFarcy (approved):

public/locales/fr/translation.json
```diff
@@ -1008,6 +1018,7 @@
       }
     },
     "usergroup": {
+      "forcedCaseName": "Group",
```
dtrucs, line 1021: In english?

src/modules/RA/Picture/components/CustomToolbar.js
```diff
@@ -65,6 +65,130 @@ const CancelButton = ({ redirect, className }) => (
...
+        isOpen={open}
+        title={translate('resources.picture.actions.accept_modal_title')}
+        content={translate('resources.picture.actions.accept_modal_content')}
+        onConfirm={handleConfirm}
+        onClose={() => setOpen(false)}
+      />
+    </>
+  );
+};
+
```
dtrucs, line 191: There are a lot of duplication for the same work.
Why don't you pass a boolean prop to display the reason or not?

src/modules/RA/Picture/components/PictureFields.js
```diff
@@ -47,6 +49,16 @@ const useStyles = makeStyles({
       height: 'auto',
       maxHeight: 'inherit',
     },
+    '& .refusal': {
+      position: 'absolute',
+      bottom: '1em',
+      width: '100%',
+      padding: '2em',
```
dtrucs, line 56: Maybe this element will go outer the width => 100% + 2em + 2em.
The safe workaround in absolute element is to not set a `width: 100%` but `right: 0; left: 0` so you can set the padding you want without side effect

dtrucs (changes requested):

public/locales/fr/translation.json
jrmi: Good point :D

src/modules/RA/Picture/components/CustomToolbar.js
jrmi: Refactored

src/modules/RA/Picture/components/PictureFields.js
jrmi: Done

dtrucs: Almost, you kept the width :D

dtrucs (approved):

src/modules/RA/Picture/components/PictureFields.js
jrmi: Done

PR #445, opened 2021-04-22
add and load sentry with given environment

submarcos:
Try to add sentry.

I don't know if we should import sentry only in case where a dsn is provided in environment

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -5,8 +5,18 @@ import './wdyr';
 import React from 'react';
 import ReactDOM from 'react-dom';
 import './index.css';
+import * as Sentry from '@sentry/react';
+import { Integrations } from '@sentry/tracing';
 import App from './App';
 import * as serviceWorker from './serviceWorker';
+
+
+Sentry.init({
+  release: process.env.npm_package_version,
+  dsn: process.env.SENTRY_DSN,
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -5,9 +5,19 @@ import './wdyr';
...
 import App from './App';
 import * as serviceWorker from './serviceWorker';
 
+
+Sentry.init({
+  release: process.env.npm_package_version,
+  dsn: process.env.SENTRY_DSN,
+  environment: process.env.SENTRY_ENVIRONMENT,
+  integrations: [new Integrations.BrowserTracing()],
+});
```
dtrucs, line 19: Maybe you can wrap sentry.init by a condition verifying the existence of `process.env.SENTRY_DSN` and in the same way for `environment` like following 

```suggestion
if(process.env.SENTRY_DSN !== undefined) {
  Sentry.init({
    release: process.env.npm_package_version,
    dsn: process.env.SENTRY_DSN,
    ...(process.env.SENTRY_ENVIRONMENT && { environment: process.env.SENTRY_ENVIRONMENT}),
    integrations: [new Integrations.BrowserTracing()],
  });
}
```

NB: Not sure if the second one is relevant; I think sentry must manage the environment if it was set to undefined

dtrucs (changes requested):

PR #460, opened 2021-04-29
Add default values for lat and lng when creating a viewpoint

vllorca:


vllorca:
image

It changes the display of the form but prevents the creation of invalid coordinate object

jrmi (approved):

PR #464, opened 2021-04-29
Delete instead of update extra geom if the value equals to null

dtrucs:


src/modules/CRUD/components/Details/PropertyItem/Edit/Edit.js
```diff
@@ -93,24 +105,35 @@ const Edit = ({
...
     );
 
     if (!isMounted.current) return;
-
-    setLoading(false);
-    if (savedFeatures.feature) {
+    if (!savedFeatures.error) {
+      getSettings(settingsEndpoint);
+      if (isDeleting) {
+        await fetchFeature(featureEndpoint, id);
```
jrmi, line 124: You need another mounted guard after this await I think.

dtrucs: Done 👍

jrmi (approved):

PR #470, opened 2021-04-30
Add relations to layer control

dtrucs:


src/modules/CRUD/views/MapPlayground/components/LayersControl/LayersControl.js
```diff
@@ -13,24 +14,74 @@ export class LayersControl extends AbstractMapControl {
...
 
+  getRelationLayerID = id => {
+    const { featureID } = this.props;
+    return `CRUD-${featureID}-relation-${id}`;
+  }
+
+  onChangeRelation = async ({ target: { value, checked } }, geojson) => {
+    const { map, getMapStyle, layers } = this.props;
+    const currentSourceAndLayerID = this.getRelationLayerID(value);
+    if (checked) {
```
jrmi, line 60: May be an early return can enhance readability here ?

src/modules/CRUD/views/MapPlayground/components/LayersControl/LayersControl.js
```diff
@@ -13,24 +14,74 @@ export class LayersControl extends AbstractMapControl {
...
+      if (!map.getSource(currentSourceAndLayerID)) {
+        let data = {};
+        try {
+          const request = await Api.request(geojson.replace('/api/', ''));
+          data = request.results;
+        } catch (e) {
```
jrmi, line 66: Not a big fan of silent catch all. Can you filter the error or at least console.log the error? Why do you have this try..catch block?

src/modules/CRUD/views/MapPlayground/components/LayersControl/LayersControl.js
```diff
@@ -59,19 +109,38 @@ export class LayersControl extends AbstractMapControl {
...
                     label={title}
                     value={id}
-                    onChange={onChange}
-                    defaultChecked={defaultChecked}
                   />
                 );
               })}
+
+              {relations
+                .sort((a, b) => a.order - b.order)
```
jrmi, line 128: Is it a good idea to put all this logic outside of render function?

src/modules/CRUD/views/MapPlayground/components/LayersControl/LayersControl.js
dtrucs: Done

src/modules/CRUD/views/MapPlayground/components/LayersControl/LayersControl.js
dtrucs: I add a toaster with the message of the API return 👍

src/modules/CRUD/views/MapPlayground/components/LayersControl/LayersControl.js
dtrucs: You're right, i refacto this logic by using a state to sanitize all the things

jrmi (approved):

PR #473, opened 2021-05-04
Avoid error raised from unmounted component

CFarcy:


vllorca (approved):

src/modules/RA/Campaign/components/CloneCampaignButton.js
```diff
@@ -31,11 +32,18 @@ const CloneCampaignButton = ({ record, ...rest }) => {
         setViewpoints(vp);
         setLoading(false);
       } catch (err) {
+        /* Error raised from unmounted component is skipped
+          to avoid setState error on unmounted component */
+        if (!isMounted) {
```
jrmi, line 37: Should also be present before the other state changes.

CFarcy: Done.

CFarcy: Done.

jrmi (approved):

PR #482, opened 2021-06-02
Do not allow new city in viewpoint creation

CFarcy:


jrmi (approved):

jrmi (changes requested):
Don't use FreeAutoComplete

jrmi (approved):

PR #485, opened 2021-10-07
Add source filter field edition with validation

jrmi:


src/modules/RA/DataLayer/components/useSourceData.js
```diff
@@ -21,10 +21,12 @@ export const useSourceData = sourceFieldName => {
     let isMounted = true;
 
     const loadSourceData = async () => {
-      const { data } = await dataProvider.getOne(RES_DATASOURCE, { id: sourceId });
-      if (!isMounted) {
+      const dataSourceResult = await dataProvider.getOne(RES_DATASOURCE, { id: sourceId });
+
+      if (!isMounted || !dataSourceResult) {
```
dtrucs, line 26: we keep the loading if the result returns nothing?

dtrucs (approved):

PR #486, opened 2021-10-13
Correctly fetch record to populate cloned source

CFarcy:


src/components/react-admin/CustomCloneButton.js
```diff
@@ -0,0 +1,78 @@
+import React, { useState, useCallback } from 'react';
+import { Link, useHistory } from 'react-router-dom';
+import {
+  useDataProvider,
+  useTranslate,
+  LinearProgress,
+  Button,
+  useResourceContext,
+  // CloneButton,
```
jrmi, line 9: I think you can remove this comment

CFarcy: done

PR #533, opened 2023-04-28
Choix de superposition des icônes

Bo-Duke:
Added boolean selector for `icon-allow-overlap`, works on the React-Admin side / visual side but nothing is connected


mabhub (approved):
LGTM
Maybe to be squashed & rebased before proper merge.

PR #544, opened 2023-06-30
Piecharts now use circles props

Bo-Duke:
See https://github.com/Terralego/terra-front/pull/347


src/modules/RA/DataLayer/components/tabs/StyleTab/Style/PiePieceStyleField.js
```diff
@@ -205,8 +206,10 @@ const PiePieceStyleField = ({ path, fields }) => {
       )}
 
       <FormControlLabel
+        style={{ marginTop: 10 }}
```
mabhub, line 209: Wouldn't a dedicated className avoid extraneous renders?

mabhub (approved):

mabhub (dismissed):

PR #551, opened 2023-08-30
Use correct status icon for source listview

CFarcy:


public/locales/fr/translation.json
```diff
@@ -395,7 +395,7 @@
         "started": "Commencée à",
         "ended": "Finie à"
       },
-      "status": "status"
+      "status": "statut"
```
submarcos, line 398:
```suggestion
      "status": "Statut"
```

PR #556, opened 2023-10-03
Add ability to declare "embed" content to be displayed as iframe

mabhub:


src/modules/RA/DataLayer/components/tabs/EmbedTab/EmbedItemInput.js
```diff
@@ -0,0 +1,101 @@
+import React from 'react';
+import { SelectInput, TextInput } from 'react-admin';
+
+import { Icon } from '@blueprintjs/core';
+
+const bpIcons = [
```
Bo-Duke, line 6: Maybe we should have this in a seperate file? A list of all usable bp icons is not a bad idea and could be used elsewhere

src/modules/RA/DataLayer/components/tabs/EmbedTab/EmbedConfigField.js
```diff
@@ -0,0 +1,46 @@
...
+  } = useInput(rest);
+
+  return (
+    <>
+      <Typography variant="h5" component="h2">
+        {translate(label)}
+      </Typography>
+
+      {error &&
+        error.length > 0 &&
```
CFarcy, line 25: What do you think of `error?.length > 0` instead?

CFarcy (approved):

PR #560, opened 2023-10-25
add sentry support from API settings

submarcos:


src/components/AppProvider/index.js
```diff
@@ -24,6 +26,29 @@ export class AppProvider extends React.Component {
       if (settings.token && !localStorage.getItem('tf:auth:token')) {
         localStorage.setItem('tf:auth:token', settings.token);
       }
+
+      if (settings.sentry.dsn !== '') {
```
Bo-Duke, line 30:
```suggestion
      if (settings.sentry?.dsn !== '') {
```

I fear that we'll get some errors if the default settings is not correctly set, is that possible? Will there always be a sentry entry in the settings?

Same applies for other PR if thats not the case, if it's the case then just ignore that

PR #574, opened 2024-01-25
Add widget tab component

Bo-Duke:


src/modules/RA/DataLayer/components/tabs/WidgetsTab/WidgetsTab.js
mabhub, line 1: It would probably be better to separate these 3 components into 3 separate files

mabhub (dismissed):

PR #601, opened 2025-10-16
Allow user to select icons from an external library

arnaudvergnet:
Based on https://github.com/Terralego/terra-admin/pull/597

Needs https://github.com/Terralego/TerraVisu/pull/969

Adds a new picker in the `Icon` form to use an external icon library as source. Those icon libraries are fetched from static files in the backend generated by a script in https://github.com/Terralego/TerraVisu/pull/969.

Users first need to choose the icon library of their choice from a list, then choose the icon. In case only one library is available, the library list is skipped.

Users can use a search bar to filter icons on their name, aliases and tags, and can also filter by full tags using the side bar.

Once an icon is selected, the icon is converted to base64 and added to the form like with the other pickers. The icon name is used to populate the `name` field and the library id is saved in the `source` field. To display the pretty library name instead of the id in the list view, the libraries `index.json` file is fetched in the list view (see https://github.com/Terralego/TerraVisu/pull/969 to understand how the backend exposes the icons).

Icon form demo:

https://github.com/user-attachments/assets/753b210e-258d-4c70-8f19-157d63caa4bb

When selecting available icons from a field (such as within a layer's style) icons are displayed sorted by source then name:

image


Bo-Duke:
Feature handled in https://github.com/Terralego/terra-admin/pull/597, closing

